PR #3, opened 2015-08-24
global: fix import of invenio_upgrader

tiborsimko:
* FIX Fixes import of invenio_upgrader in the past upgrade recipes
  following its separation into standalone package.

Signed-off-by: Tibor Simko <tibor.simko@cern.ch>


lnielsen:
:shipit:

PR #9, opened 2015-09-02
global: removal of mailcookie 

jirikuncar:
Removes dependency on *Invenio-Access* module and changes way how the access tokens are generated.

--
cc @lnielsen @hachreak

### DIFF
--- a/invenio_accounts/tokens.py
+++ b/invenio_accounts/tokens.py
@@ -1,0 +1,107 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2015 CERN.
+#
+# Invenio is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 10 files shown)


invenio_accounts/tokens.py
```diff
@@ -0,0 +1,212 @@
...
+        dt = expires_in or cfg.get('ACCOUNTS_CONFIRMLINK_EXPIRES_IN')
+
+        super(EmailConfirmationSerializer, self).__init__(
+            cfg['SECRET_KEY'],
+            expires_in=dt,
+            salt='accounts-email',
+        )
+
+
+class SecretLinkSerializer(JSONWebSignatureSerializer,
```
lnielsen: Are you using the SecretLinkSerializers at all? Otherwise I think we can just remove them.

jirikuncar: I don't think so, hence I will remove it.

tests/test_tokens.py
```diff
@@ -0,0 +1,77 @@
...
+from __future__ import absolute_import
+
+from datetime import datetime, timedelta
+
+from itsdangerous import BadData, BadSignature, JSONWebSignatureSerializer, \
+    SignatureExpired
+
+from invenio.testsuite import InvenioTestCase, make_test_suite, run_test_suite
+
+from invenio_access.tokens import EmailConfirmationSerializer, \
```
hachreak: they still exist?

PR #10, opened 2015-09-03
installation: separated access module

SamiHiltunen:
* invenio_accounts now uses the separated invenio_access module.

Signed-off-by: Sami Hiltunen sami.mikael.hiltunen@cern.ch


jirikuncar:
The dependency will be removed in #9.

Issue #2, opened 2015-08-19
accounts: create manage hooks


eamonnmag:
So that we can create users for example for inveniomanage, e.g.

```
inveniomanage user createsuperuser --u eamonn --p mypassword --e eamonn@helloworld.com
inveniomanage user createuser --u eamonn --p mypassword --e eamonn@helloworld.com
```

jirikuncar:
#### Proposal

@eamonnmag I like your proposal, I have just few consistency suggestions:

`inveniomanage accounts create [--superadmin] -u/--user <name> -e/--email <valid_email> [-p/--passoword]`

**Question**

Shall we create commands for deleting and modifying users too?

eamonnmag:
I like your option better. I was just going with how django does it. They only have 

```
 changepassword
 createsuperuser
``` 

I think that having the ability to create and modify any type of user from this tool would be very useful though.

jirikuncar:
:+1: for `changepassword`

tiborsimko:
:+1: and it'll be good to grow options later to be able to delete users, make them inactive, etc.  (E.g. merge users would be fun to have later.)

lnielsen:
:+1: for adding simple options. Especially useful for development and bootstrapping. In a running  production I would prefer to have most stuff in the UI (which naturally doesn't exclude also having it in CLI as well).

eamonnmag:
```addrole' would also be good. We need to add roles to users, and there's nothing I see on the front end side to do that yet. I may make a start on this soon, since we really need it.

jirikuncar:
@eamonnmag `addrole` should be probably part of `invenio-access` package. (cc @hachreak)

eamonnmag:
Ah, ok, I didn't know about that. Where is that package currently residing? I can't find it anywhere.

jirikuncar:
@eamonnmag created here https://github.com/inveniosoftware/invenio-access and it's waiting for https://github.com/inveniosoftware/invenio/pull/3328 to be integrated first.

eamonnmag:
Great, thanks @jirikuncar!

lnielsen:
Btw, just for reference here's the commands from Flask-Security: https://github.com/mattupstate/flask-security/blob/develop/flask_security/script.py:

- Create user
- Create role
- Add/remove role to/from user
- Activate/deactivate  user

eamonnmag:
Can we not just use this?

lnielsen:
[...]

PR #14, opened 2015-09-08
accounts: adds manage hooks

SamiHiltunen:
* NEW Adds hooks to inveniomanage for managing accounts. Implements
  functions for creating, deleting, activating and deactivating
  accounts. A hook for setting an account's password is also
  included.

* Improves code quality compliance.

Signed-off-by: Sami Hiltunen <sami.mikael.hiltunen@cern.ch>

### DIFF
--- a/invenio_accounts/manage.py
+++ b/invenio_accounts/manage.py
@@ -1,0 +1,182 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2015 CERN.
+#
+# Invenio is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Perform account operations."""
+
+from __future__ import print_function
+
+from sys import stderr
+
+from invenio.ext.script import Manager
+from invenio.ext.sqlalchemy import db
+
+from sqlalchemy.orm.exc import NoResultFound
+
+from .helpers import send_account_activation_email
+from .models import User
+from .validators import validate_email, validate_email_is_available,\
+    validate_nickname, validate_nickname_is_available, validate_password
+
+manager = Manager(usage=__doc__)
+
+
+__failed_to_delete_account = 'failed to delete the account:'
+__failed_to_activate_account = 'failed to activate the account:'
+__failed_to_deactivate_account = 'failed to deactivate the account:'
+__failed_to_set_password = 'failed to set the password:'
+__no_user_with_email = 'no user was found with email {0}'
+
+
+@manager.command
+@manager.option('--nickname', '-n', dest='nickname')
+@manager.option('--email', '-e', dest='email')
+@manager.option('--password', '-p', dest='password')
+@manager.option('--activate', '-p', dest='activate', action='store_true')
+def create(nickname, email, password, activate=False):
+    """Create a new account."""
+    try:
+        validate_nickname(nickname)
+        validate_nickname_is_available(nickname)
+        validate_email(email)
+        validate_email_is_available(email)
+        validate_password(password)
+
+        user = User()
+        user.nickname = nickname
+        user.email = email
+        user.password = password
+        if activate:
+            user.activate()
+        else:
+            user.unverify_email()
+
+        db.session.begin(subtransactions=True)
+        db.session.add(user)
+        db.session.commit()
+        db.session.commit()
+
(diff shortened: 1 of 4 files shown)


jirikuncar:
- [ ] please use `stderr` for printing error messages
- [ ] include information about user when the command is successful (see e.g. https://github.com/inveniosoftware/kwalitee/blob/master/kwalitee/cli/account.py)
- [ ] add explicit `db.begin` if you use `db.commit` and `db.rollback`

SamiHiltunen:
@jirikuncar those points should be addressed now.

invenio_accounts/manage.py
```diff
@@ -0,0 +1,136 @@
...
+        user.email = email
+        user.password = password
+        user.note = 2
+
+        db.session.begin()
+        db.session.add(user)
+        db.session.commit()
+
+        send_account_activation_email(user)
+        print(('succesfully created account %s with nickname %s, '
```
jirikuncar: can you use `.format` instead of `%`?

invenio_accounts/manage.py
```diff
@@ -0,0 +1,136 @@
...
+               'activation email has been sent') % (email, nickname))
+    except Exception as e:
+        db.session.rollback()
+        exit('failed to create a new account: %s' % e)
```
jirikuncar: `print('...', file=sys.stderr)`

jirikuncar: https://docs.python.org/2/library/constants.html?highlight=exit#exit

invenio_accounts/models.py
```diff
@@ -72,6 +75,22 @@ def __str__(self):
     """List of fields that can be updated with update_profile."""
 
     @staticmethod
+    def get_by_nickname(nickname):
+        """Get a user by their nickname."""
+        try:
+            return User.query.filter_by(nickname=nickname).one()
+        except NoResultFound:
+            raise NoResultFound('no user found with nickname %s' % nickname)
```
jirikuncar: do you really have to reraise the same exception?

invenio_accounts/models.py
```diff
@@ -159,6 +178,23 @@ def verify_email(self, force=False):
...
+    def activate(self):
+        """Activate the account."""
+        self.note = 1
+
+    def deactivate(self):
+        """Deactivate the account."""
+        self.note = 0
+
+    def unverify_email(self):
+        """
```
jirikuncar: join it with next line

jirikuncar:
@eamonnmag can you give it a try and let us know if you need something more. Thanks

jirikuncar:
@SamiHiltunen please add `(closes #2)` to commit message.

SamiHiltunen:
@jirikuncar modified the commit message

invenio_accounts/models.py
```diff
@@ -159,6 +179,20 @@ def verify_email(self, force=False):
             return True
         return False
 
+    def activate(self):
+        """Activate the account."""
+        self.note = 1
```
lnielsen: Note is a string, so it should be ``self.note = "1"`` (and same below)

invenio_accounts/manage.py
```diff
@@ -0,0 +1,140 @@
...
+    validate_nickname, validate_nickname_is_available, validate_password
+
+manager = Manager(usage=__doc__)
+
+
+@manager.command
+@manager.option('--nickname', '-n', dest='nickname')
+@manager.option('--email', '-e', dest='email')
+@manager.option('--password', '-p', dest='password')
+def create(nickname, email, password):
```
lnielsen: Can we add a flag option here ``--activate``, which will cause the account to be activated immediately and no notification send to the user.

invenio_accounts/validators.py
```diff
@@ -95,3 +101,30 @@ def validate_email(email):
...
+              x_pass=min_length))
+
+
+def validate_nickname_is_available(nickname):
+    """Check whether the nickname is taken."""
+    try:
+        User.get_by_nickname(nickname)
+    except NoResultFound:
+        return
+    raise validators.ValidationError('nickname %s is not available' % nickname)
```
jirikuncar: Please make the user facing validation error translatable (use `%(x_name)s` instead of plain `%`).

invenio_accounts/validators.py
```diff
@@ -95,3 +101,30 @@ def validate_email(email):
...
+
+
+def validate_email_is_available(email):
+    """Check whether the email is already taken."""
+    try:
+        User.get_by_email(email)
+    except NoResultFound:
+        return
+    raise validators.ValidationError('email %s is not available' % email)
```
jirikuncar: the same here (https://github.com/inveniosoftware/invenio-accounts/pull/14/files#r39503039)

invenio_accounts/models.py
```diff
@@ -72,6 +76,22 @@ def __str__(self):
     """List of fields that can be updated with update_profile."""
 
     @staticmethod
+    def get_by_nickname(nickname):
+        """Get a user by their nickname."""
+        try:
```
jirikuncar: I'm not convinced about wrapping SQLAlchemy exception. It would make sense if it was high-level API, but here we are still in model.

lnielsen: Agree. I do think it makes sense to have query methods like ``get_by_nickname`` by I would either simply raise ``NoResultFound`` or return ``None``.

SamiHiltunen:
I've made all the requested changes.

jirikuncar:
- [ ] avoid doing double `db.session.commit` (check `with db.session.begin_nested():`)

jirikuncar:
Superseded by #23

PR #36, opened 2015-11-02
accounts: fix blueprint registration

jirikuncar:
* Registers blueprint in order to load default templates.


lnielsen:
Blueprint should already be registered via the entry point: https://github.com/inveniosoftware/invenio-accounts/blob/next/setup.py#L138-L140

PR #47, opened 2015-11-12
admin: fix celery configuration

### LINKED ISSUE #34: admin: change password not working due to wrong celery app config
When you try to change the password, invenio-accounts will send a task to celery to send an email. The task however seems to be registered to the wrong Celery application, hence it tries to connect to AMQP instead of the defined value in BROKER_URL.

JavierDelgadoFernandez:
* Solves celery configuration allowing the
  execution of tasks with the correct broker.
  (closes #34)

Signed-off-by: Javier Delgado <javier.delgado.fernandez@cern.ch>

### DIFF
--- a/invenio_accounts/ext.py
+++ b/invenio_accounts/ext.py
@@ -27,7 +27,8 @@
 from __future__ import absolute_import, print_function
 
 import pkg_resources
-from flask.ext.security import Security, SQLAlchemyUserDatastore
+from flask_celeryext import FlaskCeleryExt
+from flask_security import Security, SQLAlchemyUserDatastore
 from invenio_db import db
 
 from .cli import roles as roles_cli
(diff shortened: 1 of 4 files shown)


invenio_accounts/ext.py
```diff
@@ -62,6 +63,8 @@ def init_app(self, app, use_celery=True):
             def delay_security_email(msg):
                 send_security_email.delay(msg)
 
+            FlaskCeleryExt(app)
```
jirikuncar: IMHO the extension initialization should not be here.

jirikuncar: maybe you can just assert `'invenio-celery' in app.extensions` /cc @lnielsen

lnielsen: Yes, FlaskCeleryExt is being initialized in Invenio-Celery so this would cause double initialization.

Perhaps it's just missing in the example app?

setup.py
```diff
@@ -49,6 +49,7 @@
 extras_require = {
     'celery': [
         'celery>=3.1.0',
+        'Flask-CeleryExt>=0.1.0',
```
jirikuncar: it's required only for `tests`.

PR #53, opened 2015-11-20
tests: e2e/integration tests

otron:
* Adds E2E/integration tests in `tests/e2e/`, and necessary changes to get them running locally and on Travis.
* Parametrizes the E2E tests via an environment variable and override of pytest's `pytest_generate_tests` in `tests/e2e/conftest.py`.
* Changes some existing tests to work with `pytest-flask`, which provides useful fixtures for e2e-testing.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,6 +43,7 @@ tests_require = [
     'pytest-pep8>=1.0.6',
     'pytest>=2.8.0',
     'pytest-flask>=0.10.0',
+    'selenium>=2.48.0',
     'Flask-CeleryExt>=0.1.0',
     'Flask-CLI>=0.2.1',
     'Flask-Mail>=0.9.1',
--- a/setup.py
+++ b/setup.py
@@ -42,6 +42,7 @@ tests_require = [
     'pytest-cov>=1.8.0',
     'pytest-pep8>=1.0.6',
     'pytest>=2.8.0',
+    'pytest-flask>=0.10.0',
     'Flask-CeleryExt>=0.1.0',
     'Flask-CLI>=0.2.1',
     'Flask-Mail>=0.9.1',
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -1,0 +1,99 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2015 CERN.
+#
+# Invenio is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+
+"""Pytest configuration."""
+
+from __future__ import absolute_import, print_function
+
+import os
+import shutil
+import tempfile
+
+import pytest
+from flask import Flask
+from flask_babelex import Babel
+from flask_cli import FlaskCLI
+from flask_mail import Mail
+from flask_menu import Menu
+from invenio_db import InvenioDB, db
+from selenium import webdriver
+from sqlalchemy_utils.functions import create_database, database_exists, \
+    drop_database
+
+from invenio_accounts import InvenioAccounts
+from invenio_accounts.views import blueprint
+
+
+@pytest.fixture()
+def app(request):
+    """Flask application fixture for E2E/integration/selenium tests.
+
+    Overrides the `app` fixture found in `../conftest.py`. Tests/files in
+    this folder and subfolders will see this variant of the `app` fixture.
+    """
+    instance_path = tempfile.mkdtemp()
+    app = Flask('testapp', instance_path=instance_path)
+    app.config.update(
+        ACCOUNTS_USE_CELERY=False,
+        CELERY_ALWAYS_EAGER=True,
+        CELERY_CACHE_BACKEND="memory",
+        CELERY_EAGER_PROPAGATES_EXCEPTIONS=True,
+        CELERY_RESULT_BACKEND="cache",
+        MAIL_SUPPRESS_SEND=True,
+        SECRET_KEY="CHANGE_ME",
+        SECURITY_PASSWORD_SALT="CHANGE_ME_ALSO",
+        SQLALCHEMY_DATABASE_URI=os.environ.get(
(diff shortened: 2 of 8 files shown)


setup.py
```diff
@@ -42,6 +42,7 @@
     'pytest-cov>=1.8.0',
     'pytest-pep8>=1.0.6',
     'pytest>=2.8.0',
+    'pytest-flask>=0.10.',
```
jirikuncar: `0.10.0`?

otron: Yup! My bad.

tests/e2e/conftest.py
```diff
@@ -0,0 +1,94 @@
...
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+
+"""Pytest configuration."""
+
+from __future__ import absolute_import, print_function
+
+import os
+
+from flask import Flask
```
lnielsen: Please check import sorting

otron: With `Isort`, you mean? 'Cause as it is it's checked by kwalitee :p

lnielsen: Hehe, yes with isort :-)

otron:
The actual code is ready to be reviewed. The commits still need to be rebased into, well, I don't know how many would be sensible here.

The Python 3.3/requirements=release build fails b/c the E2E tests time out when run with PhantomJS. Not sure why that happens.

otron:
Sometimes an E2E test run will timeout in a job, causing the entire build to fail.
E.g. https://travis-ci.org/inveniosoftware/invenio-accounts/jobs/94359708, which is the only failed job in the https://travis-ci.org/inveniosoftware/invenio-accounts/builds/94359695 build.

jirikuncar:
@otron maybe there is a way how py.test can re-run test which raises specific exception.

otron:
@jirikuncar it doesn't raise any particular exception though. Travis kills the job after 10 minutes.

tests/e2e/conftest.py
```diff
@@ -0,0 +1,109 @@
...
+
+@pytest.fixture()
+def env_browser(request):
+    """Fixture for a webdriver instance of the browser specified by the
+    `WEBDRIVER_BROWSER` environment variable."""
+
+    # Create instance of webdriver.`request.param`()
+    browser = getattr(webdriver, request.param)()
+    # If the test fails it's be useful to know which browser was used.
+    print("Webdriver browser=", request.param)
```
jirikuncar: using `print` is not the best way to to detect fail - you should have this information in traceback context.

otron: You are right. Turns out it's also not required, as `pytest` prints the value of `env_browser` as a part of the failure information:
```
live_server = <LiveServer listening at http://localhost:53179>
env_browser = <selenium.webdriver.firefox.webdriver.WebDriver (session="aed62c3e-8180-0341-b571-eb261a5d987e")>

    def test_user_registration(live_server, env_browser):
        """E2E user registration and login test."""
        browser = env_browser
>       assert False
```

otron:
Travis build fails b/c the e2e tests timeout on random jobs :(

tests/test_utils.py
```diff
@@ -93,16 +95,20 @@ def test_client_authenticated(app):
...
 
     Demonstrates basic usage and context requirements."""
     ext = InvenioAccounts(app)
     email = 'test@test.org'
     password = '1234'
 
-    # Can't access the datastore outside of an application context
-    with pytest.raises(RuntimeError):
-        testutils.create_test_user(email, password)
+    # If pytest-flask isn't available we can't access the datastore outside of
```
jirikuncar: since we `pytest-flask` is part of test dependencies i'm not sure why do we need to keep it here.

otron: Only reason I can think of is as a helpful reminder to future developers who might not be aware of why this repo's tests are different than the ones in repos not using pytest-flask. But this probably isn't the place to include a tidbit like that.

jirikuncar: Testing section in docs might be a better place.

tests/e2e/conftest.py
```diff
@@ -0,0 +1,120 @@
...
+    Babel(app)
+    Mail(app)
+    InvenioDB(app)
+    InvenioAccounts(app)
+    app.register_blueprint(blueprint)
+
+    with app.app_context():
+        if not database_exists(str(db.engine.url)):
+            create_database(str(db.engine.url))
+        db.drop_all()
```
jirikuncar: it should not be necessary

otron: Which part?

jirikuncar: `db.drop_all()`

otron: While it _shouldn't_ be necessary, I feel it's one of those "better safe than having tests inexplicably fail and then pass on the next run because of some weirdness left in the database"-situations.

jirikuncar: It means that you haven't cleaned your db on tear down and your test MUST necessarily fail then.

otron: Or the process could have been interrupted for some reason, causing the
teardown stage to be skipped.

lnielsen: It's useful when you kill tests with ctrl-c during development so you don't have to clean up the DB manually in those cases.

jirikuncar: IMHO it's not good idea. If you kill the tests with `ctrl-c` then you can run `db drop --yes-i-know`.

- normal runtime of tests is longer
- you might not discover that something went wrong after your tests

jirikuncar: > It's useful when you kill tests with ctrl-c during development so you don't have to clean up the DB manually in those cases.

Lazy developer vs. clean code ... I'm personally against this practice, but if the majority wants to do it I'm not going to block it. I hope I gave enough good reasons so you can re-consider your decision.

lnielsen: Well, it's all a matter of perspective. "Lazy developer vs. clean code" could just as well be put as "Complex usage vs. User friendly usage" depending on how you look at it ;-)

jirikuncar: PyTest have you covered - it automatically calls `request.finalizers` on `KeyboardInterupt` :wink: Hence, it's even more important that you do correct clean-up. It cleans when the test fails too, hence you just need to re-run the tests if your environment is in dirty state.

otron: I'm inclined to agree with @jirikuncar that it isn't necessary to drop the db during initialization, as the finalizers are run on non-exceptional interrupts and should clean out the db.

However this currently exists in the `master` branch in [`tests/conftest.py`](https://github.com/inveniosoftware/invenio-accounts/blob/master/tests/conftest.py#L78) so I'd rather keep it here to minimize the differences between the two `app` fixtures.

otron:
Changed this back to WIP while I look for a fix for the Travis timeouts.

otron:
Update on the Travis timeouts:
1. Added a timeout to the `env_browser` fixture.
 This revealed that the tests were timing out when run with PhantomJS.
 Supported by [job #331.2](https://travis-ci.org/inveniosoftware/invenio-accounts/jobs/97638304), [job #331.8](https://travis-ci.org/inveniosoftware/invenio-accounts/jobs/97638315) and [job #332.26](https://travis-ci.org/inveniosoftware/invenio-accounts/jobs/97659282)

2. Removed PhantomJS from the list of browsers in `.travis.yml`.
  While [build #333](https://travis-ci.org/inveniosoftware/invenio-accounts/builds/97663340) and [build #335](https://travis-ci.org/inveniosoftware/invenio-accounts/builds/100080238) didn't fail, this doesn't prove that the problem has been fixed as we don't have data for all previously timed out jobs.

Issue #44, opened 2015-11-06
tests: add tests for user deactivation


otron:
Need test(s) that verify that a user, once deactivated, immediately has their session(s) invalidated and are unable to log in.

Test description:
1. Create user X
2. log in as X
3. deactivate user X
4. try to do _things_ as user X. Result should be the same as trying to do so as any anonymous/unauthenticated user

otron:
How do we deactivate a user in `invenio-accounts`? Do we use `flask_security.datastore.deactive_user(user)`[0]?

[0]: https://pythonhosted.org/Flask-Security/api.html?highlight=deactivate#flask_security.datastore.UserDatastore.deactivate_user

lnielsen:
The admin should provide a filter action to deactivate multiple users at once, which should use the Flask-Security datastore.deactivate_user under the hood.

otron:
Admin as in `invenio-admin`?

lnielsen:
In #33

jirikuncar:
Solved as part of #33.

Issue #52, opened 2015-11-18
global: session activity feature


lnielsen:
We need the ability to destroy a valid session without the user making a request (this is useful to e.g. ban users immediately #44). Also, this can be used to allow users to see a list of their active sessions (e.g. from multiple devices, and logout a specific device). This is only really possible when sessions are stored server-side (i.e #51). We also need to associate a session (for a logged in user) with a user id (Flask-Session/KVSession use uuids as session ids).

- [x] Model: user, session, timestamp (UTC), country, browser, os (deduced from ip and user agent)
- [x] Connect to Flask-Login's `user_logged_in` signal to store the session activity when a user logins.
- [x] Celery cron task to clean up session activity table.
- [x] Account settings item "Security", with a panel that list Sessions (see e.g. DropBox as an example). Allow a user to remove a session plus "Log me out of all devices". I would also include "Change password" on the same page.
- [x] Model: +country, browser, browser version, os (deduced from ip and user agent)
- [x] Configuration: Allow 1) tracking only user/session/timestamp and not country, browser and os 2) completely switching off the session activity feature.
- [x] Admin interface (needs very restrictive permissions). Browse, remove, filter by user. (related to #33)

jirikuncar:
@lnielsen I'm not sure if all tasks were done in https://github.com/inveniosoftware/invenio-accounts/pull/65. Shall we re-open and mark tasks that has been done?

lnielsen:
@hachreak We should as minimum add the celery clean up script and "security" so user have full control of their session data.

lnielsen:
Closed by 1ade848

Issue #51, opened 2015-11-17
session: integrate Flask-KVSession with Redis backend


jirikuncar:
- [ ] `accounts_session_activity` table
- [ ] `user_logged_in` signal for populating activity table

/ cc @lnielsen

PR #58, opened 2015-11-24
WIP tests: session removal after user deactivation

otron:
* addresses #44 
* Currently fails as the user's session is not invalidated upon said user's deactivation.
* Won't pass until #51 and #52 are taken care of

### DIFF
--- a/invenio_accounts/sessions.py
+++ b/invenio_accounts/sessions.py
@@ -89,3 +89,15 @@ def delete_session(sid_s):
     _datastore.db.session.delete(sessionactivity)
     _datastore.commit()
     return 1
+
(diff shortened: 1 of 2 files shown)


jirikuncar:
@otron is it still relevant?

otron:
@jirikuncar yes, deactivating a user should invalidate all their sessions. This was not implemented as a part of #65, so unless someone else has addressed/resolved #44 this is still relevant.

jirikuncar:
@otron can you at least rebase this branch so one can easily start working on it? Thanks

otron:
@jirikuncar It is done.

PR #61, opened 2015-11-24
installation: fix for ipaddr in wheel on PY3

### LINKED ISSUE #57: setup: wheel not universal
The wheel built on Python 2.7 and uploaded on PyPI cannot be installed on Python 3 due to dependency on ``ipaddr`` (which in setup.py is only added in 2.7, but the wheel doesn't care)

jirikuncar:
* Makes a wheel generated on Python 2.7 universal for Python 3.3+.
  (closes #57)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,8 +30,6 @@ import sys
 from setuptools import find_packages, setup
 from setuptools.command.test import test as TestCommand
 
-PY3 = sys.version_info[0] == 3
-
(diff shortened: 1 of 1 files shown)


jirikuncar:
@lnielsen the real issue was here https://github.com/inveniosoftware/invenio-accounts/pull/61/files#diff-2eeaed663bd0d25b7e608891384b7298R63

PR #65, opened 2015-12-02
sessions: backend sessions and user activity

### LINKED ISSUE #51: session: integrate Flask-KVSession with Redis backend

otron:
Not at all anything like _backyard sessions_.

resolves #51.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,7 @@ cache:
 services:
   - mysql
   - postgresql
+  - redis
 
 env:
   - REQUIREMENTS=lowest E2E_WEBDRIVER_BROWSERS="" EXTRAS=all,sqlite SQLALCHEMY_DATABASE_URI="sqlite:///test.db"
--- a/invenio_accounts/ext.py
+++ b/invenio_accounts/ext.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2015 CERN.
+# Copyright (C) 2015, 2016 CERN.
 #
 # Invenio is free software; you can redistribute it
 # and/or modify it under the terms of the GNU General Public License as
@@ -27,33 +27,50 @@
 from __future__ import absolute_import, print_function
 
 import pkg_resources
+from flask_kvsession import KVSessionExtension
 from flask_security import Security, SQLAlchemyUserDatastore
 from invenio_db import db
+from flask_login import user_logged_in
 
 from .cli import roles as roles_cli
 from .cli import users as users_cli
 from .models import Role, User
+from .sessions import login_listener
 
 
 class InvenioAccounts(object):
     """Invenio-Accounts extension."""
 
-    def __init__(self, app=None):
+    def __init__(self, app=None, sessionstore=None):
         """Extension initialization."""
         self.security = Security()
         self.datastore = None
         if app:
-            self.init_app(app)
+            self.init_app(app, sessionstore=sessionstore)
 
-    def init_app(self, app, use_celery=True):
-        """Flask application initialization."""
+    def init_app(self, app, use_celery=True, sessionstore=None):
+        """Flask application initialization.
+
+        :param sessionstore: store for sessions. Passed to
+            ``flask-kvsession``. Defaults to redis.
+        """
         self.init_config(app)
 
         # Create user datastore
         self.datastore = SQLAlchemyUserDatastore(db, User, Role)
 
+        # Create sessionstore
+        if sessionstore is None:
+            import redis
+            from simplekv.memory.redisstore import RedisStore
+            # TODO: use default from config?
+            sessionstore = RedisStore(redis.StrictRedis())
+        self.sessionstore = sessionstore
+        user_logged_in.connect(login_listener, app)
+
         # Initialize extension.
         state = self.security.init_app(app, datastore=self.datastore)
+        self.kvsession_extension = KVSessionExtension(self.sessionstore, app)
 
         if app.config['ACCOUNTS_USE_CELERY']:
             from invenio_accounts.tasks import send_security_email
--- a/invenio_accounts/models.py
+++ b/invenio_accounts/models.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2015 CERN.
+# Copyright (C) 2015, 2016 CERN.
 #
 # Invenio is free software; you can redistribute it
 # and/or modify it under the terms of the GNU General Public License as
@@ -29,7 +29,7 @@ from __future__ import absolute_import, print_function
 from flask_security import RoleMixin, UserMixin
 from invenio_db import db
 from sqlalchemy.orm import validates
-from sqlalchemy_utils import IPAddressType
+from sqlalchemy_utils import IPAddressType, Timestamp
 
 userrole = db.Table(
     'accounts_userrole',
@@ -87,3 +87,25 @@ class User(db.Model, UserMixin):
         if value == 'untrackable':  # pragma: no cover
             value = None
         return value
+
+
+class SessionActivity(db.Model, Timestamp):
+    """User Session Activity model.
+
+    Instances of this model correspond to a session belonging to a user.
+    """
+
+    __tablename__ = "accounts_user_session_activity"
+
+    id = db.Column(db.Integer, primary_key=True)
+
+    user_id = db.Column(db.Integer, db.ForeignKey(User.id))
+    """ID of user to whom this session belongs."""
+
+    sid_s = db.Column(db.String(255), unique=True, index=True)
+    """Serialized Session ID. Used as the session's key in the kv-session
+    store employed by `flask-kvsession`.
+    Named here as it is in `flask-kvsession` to avoid confusion.
+    """
+
+    user = db.relationship(User, backref='active_sessions')
--- a/invenio_accounts/sessions.py
+++ b/invenio_accounts/sessions.py
@@ -1,0 +1,91 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2015 CERN.
+#
+# Invenio is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Backend session management for accounts.
+
+The actual backend sessions are provided by ``flask-kvsession``.
+This module provides management functionality and populates the
+:class:`invenio_accounts.models.SessionActivity` table as new sessions are
+created.
+"""
+
+from __future__ import absolute_import, print_function
+
+import flask
+from flask_login import user_logged_in
+from sqlalchemy.exc import IntegrityError
+from werkzeug.local import LocalProxy
+
+from invenio_accounts.models import SessionActivity
+
+_datastore = LocalProxy(lambda: flask.current_app.
+                        extensions['invenio-accounts'].datastore)
+
+_sessionstore = LocalProxy(lambda: flask.current_app.
+                           extensions['invenio-accounts'].sessionstore)
+
+
+def add_session(session=None):
+    r"""Add a session to the SessionActivity table.
+
+    :param session: Flask Session object to add. If None, ``flask.session``
+        is used. The object is expected to have a dictionary entry named
+        ``"user_id"`` and a field ``sid_s``
+    """
+    user_id, sid_s = session['user_id'], session.sid_s
+    session_activity = SessionActivity(user_id=user_id, sid_s=sid_s)
+    try:
+        _datastore.db.session.add(session_activity)
+        _datastore.commit()
+    except IntegrityError:
+        _datastore.db.session.rollback()
+
+
+def login_listener(app, user):
+    """Connect to the user_logged_in signal for table population."""
+    @flask.after_this_request
+    def populate_session_activity(response):
+        """Add the current session to the SessionActivity table."""
+        # Have to save the session so that the sid_s gets generated.
+        app.session_interface.save_session(app, flask.session, response)
+        add_session(flask.session)
+        return response
+
+
+def delete_session(sid_s):
+    """Delete entries in the data- and kvsessionstore with the given sid_s.
+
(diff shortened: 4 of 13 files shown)


.travis.yml
```diff
@@ -67,6 +67,12 @@ install:
   - "travis_retry pip install -r .travis-${REQUIREMENTS}-requirements.txt"
   - "travis_retry pip install -e .[${EXTRAS}]"
 
+before_script:
+  - "export DISPLAY=:99.0" #set up xvfb, req'd for non-phantomJS webdrivers
+  - "sh -e /etc/init.d/xvfb start"
+  - "export E2E_WEBDRIVER_BROWSERS='Firefox PhantomJS'"
+  - "redis-server &"
```
jirikuncar:
```
services
...
  - redis
```
will do the job better

tests/test_utils.py
```diff
@@ -93,16 +95,20 @@ def test_client_authenticated(app):
...
     email = 'test@test.org'
     password = '1234'
 
-    # Can't access the datastore outside of an application context
-    with pytest.raises(RuntimeError):
-        testutils.create_test_user(email, password)
+    # If pytest-flask isn't available we can't access the datastore outside of
+    # an application context
+    try:
+        pkg_resources.get_distribution('pytest-flask')
```
jirikuncar: then we should add `pytest-flask` to `tests_require`

otron: It already is. This was left in as an example of its effect.

.travis.yml
otron: Thanks!

tests/e2e/conftest.py
```diff
@@ -0,0 +1,110 @@
...
+
+@pytest.fixture()
+def ff_browser(request):
+    """Pytest fixture for a selenium/webdriver instance with Firefox."""
+    ff = webdriver.Firefox()
+    request.addfinalizer(lambda: ff.quit())
+    return ff
+
+
+@pytest.fixture()
```
jirikuncar: ### Proposal

Remove `run-e2etests.py` and use generated tests.

```python
# content of conftest.py
import pytest

def pytest_generate_tests(metafunc):
    if 'env_browser' in metafunc.fixturenames:
        browsers = os.environ.get('E2E_WEBDRIVER_BROWSERS', 'firefox').split()
        metafunc.parametrize("env_browser", browsers, indirect=True)

@pytest.fixture
def env_browser(request):
    # Create instance of webdriver.`request.param`()
    browser = getattr(webdriver, request.param)()
    # Add finalizer to quit the webdriver instance
    request.addfinalizer(lambda: browser.quit())
    return browser
```

invenio_accounts/ext.py
```diff
@@ -52,8 +55,12 @@ def init_app(self, app, use_celery=True):
         # Create user datastore
         self.datastore = SQLAlchemyUserDatastore(db, User, Role)
 
+        # Create session store
+        self.sessionstore = RedisStore(redis.StrictRedis())
```
jirikuncar: this should be configurable at least via keyword arguments.

jirikuncar: this should be configurable at least by using keyword arguments.

tests/e2e/conftest.py
otron: Nice! Got it working locally, have pushed the changes to the PR branch for the e2e tests (https://github.com/inveniosoftware/invenio-accounts/pull/53)

invenio_accounts/ext.py
otron: to `__init__` or `init_app`?

jirikuncar: both

invenio_accounts/models.py
```diff
@@ -78,6 +80,10 @@ class User(db.Model, UserMixin):
     roles = db.relationship('Role', secondary=userrole,
                             backref=db.backref('users', lazy='dynamic'))
 
+    active_sessions = db.relationship('SessionActivity', backref='user')
```
jirikuncar: it's better to define it on `SessionActivity` model  as `user = db.relationship(User, backref='active_sessions')` (please note that you can use the class directly not its string name).

invenio_accounts/models.py
```diff
@@ -87,3 +93,14 @@ def validate_ip(self, key, value):
...
         return value
+
+
+class SessionActivity(db.Model):
+    """User Session Activity model."""
+
+    __tablename__ = "accounts_user_session_activity"
+
+    id = db.Column(db.Integer, primary_key=True)
+    user_id = db.Column(db.Integer, db.ForeignKey('accounts_user.id'))
```
jirikuncar: `user_id = db.Column(db.Integer, db.ForeignKey(User.id))`

invenio_accounts/models.py
```diff
@@ -87,3 +93,14 @@ def validate_ip(self, key, value):
...
+    sid_s = db.Column(db.String(255), unique=True)
+    timestamp = db.Column(db.DateTime, default=datetime.datetime.utcnow)
```
jirikuncar: you can use `Timestamp` mixin from SQLAlchemy-Utils.

invenio_accounts/sessions.py
```diff
@@ -0,0 +1,83 @@
...
+
+_datastore = LocalProxy(lambda: flask.current_app.
+                        extensions['invenio-accounts'].datastore)
+
+
+def add_session(user_id, sid_s):
+    """Adds the session to the SessionActivity table.
+
+    Remember to save the session via `app.session_interface.save_session`
+    first."""
```
jirikuncar:
```python
first.
"""
```

tests/test_sessions.py
```diff
@@ -0,0 +1,209 @@
...
+        # from the first?
+        # pdb.set_trace()
+        testutils.login_user_via_view(client, user_2.email,
+                                      user_2.password_plaintext)
+        assert len(app.kvsession_store.keys()) == 1
+        assert app.kvsession_store.keys()[0] == kvsession_key
+        assert True
+
+
+def get_cookie_from_client(client, sid_s=None):
```
jirikuncar: This function can be useful in `testutils`.

tests/test_sessions.py
```diff
@@ -0,0 +1,209 @@
...
+    """Extracts the cookie from the client's cookie jar matching the given
+    sid_s value. If sid_s is none, it returns the first cookie."""
+    for c in client.cookie_jar:
+        if sid_s is None or sid_s in c.value:
+            return c
+
+
+def get_sid_s_from_cookie(cookie):
```
jirikuncar: It should probably go to `invenio_accounts/sessions.py`.

tests/test_sessions.py
```diff
@@ -0,0 +1,209 @@
...
+                                      user.password_plaintext)
+        assert len(app.kvsession_store.keys()) == 1
+
+        kvsession_key = app.kvsession_store.keys()[0]
+        sesh = app.kvsession_store.get(kvsession_key)
+        cookie = get_cookie_from_client(client, kvsession_key)
+        cookie_sid_s = get_sid_s_from_cookie(cookie)
+        assert kvsession_key == cookie_sid_s
+
+        import pdb
```
jirikuncar: btw you can take advantage to `py.test tests/test_session.py --pdb` or if you need debugger `import pytest; pytest.set_trace()` can do the job too.

tests/test_sessions.py
otron: `testutils.py` is more befitting it: It complements `get_cookie_from_client`, which should only work for instances of Flask's `test_client`.

invenio_accounts/models.py
```diff
@@ -87,3 +89,17 @@ def validate_ip(self, key, value):
...
+class SessionActivity(db.Model, Timestamp):
+    """User Session Activity model."""
+
+    __tablename__ = "accounts_user_session_activity"
+
+    id = db.Column(db.Integer, primary_key=True)
+
+    user_id = db.Column(db.Integer, db.ForeignKey(User.id))
+
+    sid_s = db.Column(db.String(255), unique=True, index=True)
```
lnielsen: What does the last ``s`` stands for in ``sid_s``?

invenio_accounts/sessions.py
```diff
@@ -0,0 +1,88 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Session stuff."""
```
lnielsen: Please use a better description than ``stuff`` ;-)

invenio_accounts/sessions.py
```diff
@@ -0,0 +1,88 @@
...
+
+def login_listener(app, user):
+    """Connect to the user_logged_in signal to populate the SessionActivity
+    table when a user logs in."""
+
+    @flask.after_this_request
+    def populate_session_activity(response):
+        """Adds the current session to the SessionActivity table."""
+        # Have to save the session so that the sid_s gets generated.
+        app.session_interface.save_session(app, flask.session, response)
```
lnielsen: Does this mean that there's several ``save_session`` calls?

invenio_accounts/sessions.py
```diff
@@ -0,0 +1,88 @@
...
+    """Adds the session to the SessionActivity table.
+
+    Remember to save the session via `app.session_interface.save_session`
+    first.
+    """
+    sact = SessionActivity(user_id=user_id, sid_s=sid_s)
+    try:
+        _datastore.db.session.add(sact)
+        _datastore.commit()
+    except IntegrityError as error:
```
lnielsen: ``except IntegrityError:``

invenio_accounts/sessions.py
```diff
@@ -0,0 +1,88 @@
...
+        # Have to save the session so that the sid_s gets generated.
+        app.session_interface.save_session(app, flask.session, response)
+        add_session(user.id, flask.session.sid_s)
+        return response
+
+
+def logout_listener(app, user):
+    """Connect to the user_logged_out signal?"""
+
+    save_sesh_sid_s = flask.session.sid_s
```
lnielsen: What does the abbreviations stand for?

invenio_accounts/sessions.py
otron: From what I can tell from Flask's source code, `save_session` is called once for each response processed by the application. So yes, there's likely to be several.
These will all go to `flask_kvsession/__init__.py:KVSessionInterface:save_session`, which 1) doesn't do anything unless the session's `modified` field is `True` and 2) won't create a new session entry in the sessionstore unless the session's `sid_s` has been deleted.

How come?

invenio_accounts/models.py
jirikuncar: I would go for `session_id`

otron: Possibly the whole thing means `session id serialized` (the last `s` standing for `serialized`), as the `sid_s` value is unserialized to produce the actual session.
It is, however, _not_ the session's id and naming the field `session_id` would be confusing (looking through the code you'd then see that the value stored here as `session_id` is referred to as `sid_s` in `flask_kvsession`, and said value gets unserialized to produce the session object which contains the session id).

The field contains the session's key value as stored in `flask_kvsession`'s sessionstore.
I've named this field `sid_s` so that it corresponds with the naming used in `flask_kvsession` to avoid confusion.

invenio_accounts/sessions.py
otron: "sesh" is slang for "session". 
`sid_s` I _think_ is an abbreviation for `SessionID_serialized`. Never thought about it much, I just kept the naming convention found in `flask_kvsession`

invenio_accounts/models.py
jirikuncar: Does it contain the key to `flask_kvsession` storage or whole session? Btw adding docstrings to the columns in the new model would reduce the need of this discussion.

otron: Both -- unserializing the `sid_s` produces the session.
Sure, I can add some docstrings. I didn't know any of this when I created the model. It's good that you're pointing out stuff that's not immediately apparent, as I'd have a hard time doing the same.

invenio_accounts/sessions.py
jirikuncar: > I just kept the naming convention used in `flask_kvsession`

Please try to be closer to naming conventions of `inveniosoftware` projects by using easily understandable names without slang expressions. Thank you

invenio_accounts/models.py
otron: @jirikuncar: docstrings as in sphinx docstrings for the model/Class, or just some inline comments?

jirikuncar: for properties ... e.g. https://github.com/inveniosoftware/invenio-pidstore/blob/master/invenio_pidstore/models.py#L130

invenio_accounts/ext.py
```diff
@@ -27,33 +27,49 @@
...
+        """
         self.init_config(app)
 
         # Create user datastore
         self.datastore = SQLAlchemyUserDatastore(db, User, Role)
 
+        # Create session store
+        if sessionstore is None:
+            # TODO: use default from config?
+            sessionstore = RedisStore(redis.StrictRedis())
```
lnielsen: Move Redis imports to here, to avoid importing redis if a different sessionstore is provided.

invenio_accounts/sessions.py
```diff
@@ -0,0 +1,94 @@
...
+def logout_listener(app, user):
+    """Connect to the user_logged_out signal."""
+    save_sesh_sid_s = flask.session.sid_s
+
+    @flask.after_this_request
+    def remove_session_activity(response):
+        """Responsible for deletion of the SessionActivity entry."""
+        # Oh snap, will we know which session just got deleted?
+        import pdb
+        # pdb.set_trace()
```
lnielsen: .

invenio_accounts/sessions.py
```diff
@@ -0,0 +1,91 @@
...
+    :returns: 1 if deletion was successful.
+    """
+    # Remove entries from sessionstore
+    _sessionstore.delete(sid_s)
+    # Find and remove the corresponding SessionActivity entry
+    sessionactivity = _datastore.db.session.query(SessionActivity).\
+        filter_by(sid_s=sid_s).first()
+    _datastore.db.session.delete(sessionactivity)
+    _datastore.commit()
+    return 1
```
jirikuncar: `1` vs. `True`?

invenio_accounts/testutils.py
```diff
@@ -63,11 +66,25 @@ def create_test_user(email='test@test.org',
     return user
 
 
-def login_user_via_view(client, email, password, login_url=None):
-    """Attempt to log the given user in via the 'login' view on the client.
+def login_user_via_view(client, email=None, password=None, user=None,
```
jirikuncar: I'm sorry but I'm not a big fan of adding `user` to the API. Can we instead create a helper function in tests that sets email and password?

invenio_accounts/sessions.py
otron: `_sessionstore.delete` returns 1 if deletion was successful,
`_datastore.db.session.delete` returns nothing if deletion was succesful.

Which I guess was why I had this return 1. Though might it just as well not return anything, seeing as how an error will be thrown if either deletion fails.

jirikuncar: I see another problematic part with `commit`. `SessionActivity` is not managed by Flask-Security but by Invenio-Accounts. Hence we should use nested transaction and make sure that the commit is called at the end of request.

PR #71, opened 2015-12-07
tests: default user password fix

otron:
Passing no args to `testutils.py:create_test_user` created a user with the empty string as its password. Logging in doesn't work with such a user. This wasn't caught as all of the tests for `create_test_user` explicitly passed email and password.
The PR contains a test for the default values of `create_test_user`, and changes the default value of its `password` kwarg to a non-empty string.

### DIFF
--- a/invenio_accounts/testutils.py
+++ b/invenio_accounts/testutils.py
@@ -38,7 +38,7 @@ _datastore = LocalProxy(lambda: current_app.extensions['security'].datastore)
 
 
 def create_test_user(email='test@test.org',
-                     password='', **kwargs):
+                     password='123456', **kwargs):
     """Create a user in the datastore, bypassing the registration process.
 
     Returns the created user model object instance, with the plaintext password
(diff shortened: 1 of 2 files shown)


invenio_accounts/testutils.py
```diff
@@ -38,7 +38,7 @@
 
 
 def create_test_user(email='test@test.org',
-                     password='', **kwargs):
+                     password='cannot_be_empty', **kwargs):
```
jirikuncar: usually we add `123456` in tests

tests/test_utils.py
```diff
@@ -124,6 +124,20 @@ def test_create_test_user(app):
         # create a user with a duplicate email.
 
 
+def test_create_test_user_defaults(app):
+    """Test the default values for testutils.py:create_test_user()"""
```
jirikuncar: `"""Test the default values for testutils.py:create_test_user."""`

PR #78, opened 2015-12-16
templates: addition of a hook in login_user

JavierDelgadoFernandez:
* Adds an empty block in login_user.html allowing the insertion of
  custom code before the login fields.

Signed-off-by: Javier Delgado <javier.delgado.fernandez@cern.ch>

### DIFF
--- a/invenio_accounts/templates/invenio_accounts/login_user.html
+++ b/invenio_accounts/templates/invenio_accounts/login_user.html
@@ -34,6 +34,7 @@
       <h3 class="text-center panel-free-title">{{_('Log in to account') }}</h3>
       {%- endblock form_header %}
       {%- block form_outer %}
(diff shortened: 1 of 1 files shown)


invenio_accounts/templates/invenio_accounts/login_user.html
```diff
@@ -34,6 +34,7 @@
       <h3 class="text-center panel-free-title">{{_('Log in to account') }}</h3>
       {%- endblock form_header %}
       {%- block form_outer %}
+      {% block block_before_login_form %}{% endblock block_before_login_form %}
```
jirikuncar: you don't really need it. you can override `form_outer`:

```jinja2
{% block form_outer %}
  do something before login form
  {{ super() }}
{% endblock %}
```

--
http://jinja.pocoo.org/docs/dev/templates/#super-blocks

PR #85, opened 2016-01-14
tests: respect ACCOUNTS_USE_CELERY configuration

### LINKED ISSUE #82: celery: ACCOUNTS_USE_CELERY should be configurable and not overridden.
At present, there is no ability to override this config option since this line in ext always sets the variable depending on the mode the app is running in.

https://github.com/inveniosoftware/invenio-accounts/blob/master/invenio_accounts/ext.py#L76

jirikuncar:
* Checks that value in ACCOUNTS_USE_CELERY configuration is respected.
  (closes #82)

### DIFF
--- a/tests/test_invenio_accounts.py
+++ b/tests/test_invenio_accounts.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2015 CERN.
+# Copyright (C) 2015, 2016 CERN.
 #
 # Invenio is free software; you can redistribute it
 # and/or modify it under the terms of the GNU General Public License as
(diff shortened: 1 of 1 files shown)


eamonnmag:
:shipit:

invenio_accounts/tasks.py
```diff
@@ -30,7 +30,7 @@
 from flask import current_app
 
 
-@shared_task
+@shared_task(serializer='pickle')
```
lnielsen: We shouldn't be using ``pickle`` as serialiser. It's insecure and will be disabled by default (http://docs.celeryproject.org/en/latest/whatsnew-3.1.html#last-version-to-enable-pickle-by-default).

PR #86, opened 2016-01-15
cli: unify arguments and options

### LINKED ISSUE #56: cli: unify arguments and options
- [ ] user email should always be `-e/--email`
- [ ] should role name in `roles` be repeatable? if not then it would be better to use `argument` name instead of `-r` option

/cc @lnielsen

hachreak:
* Uses argument instead of option for required parameters. (closes #56)

### DIFF
--- a/examples/app.py
+++ b/examples/app.py
@@ -51,8 +51,8 @@ Create a user:
 
 .. code-block:: console
 
-   $ flask -a app.py users create -e info@invenio-software.org -a
-   $ flask -a app.py users activate -u info@invenio-software.org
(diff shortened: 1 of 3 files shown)


jirikuncar:
@hachreak it looks good. Please amend also examples https://github.com/inveniosoftware/invenio-accounts/blob/master/examples/app.py#L54-L55 and check Invenio installation recipe that @tiborsimko has added recently.

hachreak:
I found this line to update: https://github.com/inveniosoftware/invenio/blob/master/scripts%2Fpopulate-instance.sh#L108-L111
@tiborsimko do you know if there are other places that I should update?

PR #88, opened 2016-01-17
tests: separate E2E test build matrix on Travis

otron:
Not sure about the co-authored/signed-off-by attribution -- I've basically copied what Lars did in Zenodo.

This commit increases the Travis job count by 4 (one per Python version), meaning the total job count per build is now 40.
The E2E tests are excluded from 36 of the jobs (which now only run the unit tests), and run once in each of the four new jobs.
Does this reduce the overall build time?
- Old build time: `J * (tu + te) + JS`
- New build time: `S * (J + K) + J * tu + J * te)
- `J` is the number of jobs per build before this commit, `K` is the number of jobs added by this commit, `S` is the startup/setup time of a Travis job, `tu` is the avg. run time of the unit tests, `te` is the avg. run time of the E2E tests.
- If `J * te > K * (S + te)` (it takes less time to set up `K` travis jobs and run the E2E tests `K` times than it does to run the E2E tests `J` times), this commit reduces the total build time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,15 +40,16 @@ services:
   - postgresql
 
 env:
-  - REQUIREMENTS=lowest EXTRAS=all,sqlite SQLALCHEMY_DATABASE_URI="sqlite:///test.db"
-  - REQUIREMENTS=lowest EXTRAS=all,mysql SQLALCHEMY_DATABASE_URI="mysql+pymysql://travis@localhost:3306/invenio"
-  - REQUIREMENTS=lowest EXTRAS=all,postgresql SQLALCHEMY_DATABASE_URI="postgresql+psycopg2://postgres@localhost:5432/invenio"
-  - REQUIREMENTS=release EXTRAS=all,sqlite SQLALCHEMY_DATABASE_URI="sqlite:///test.db"
-  - REQUIREMENTS=release EXTRAS=all,mysql SQLALCHEMY_DATABASE_URI="mysql+pymysql://travis@localhost:3306/invenio"
-  - REQUIREMENTS=release EXTRAS=all,postgresql SQLALCHEMY_DATABASE_URI="postgresql+psycopg2://postgres@localhost:5432/invenio"
-  - REQUIREMENTS=devel EXTRAS=all,sqlite SQLALCHEMY_DATABASE_URI="sqlite:///test.db"
-  - REQUIREMENTS=devel EXTRAS=all,mysql SQLALCHEMY_DATABASE_URI="mysql+pymysql://travis@localhost:3306/invenio"
-  - REQUIREMENTS=devel EXTRAS=all,postgresql SQLALCHEMY_DATABASE_URI="postgresql+psycopg2://postgres@localhost:5432/invenio"
+  - REQUIREMENTS=devel E2E="yes" EXTRAS=all,postgresql SQLALCHEMY_DATABASE_URI="postgresql+psycopg2://postgres@localhost:5432/invenio"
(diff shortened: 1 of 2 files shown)


otron:
From comparing the build times PR with others, having dedicated job(s) for the E2E tests doesn't seem to reduce the build time by much.
Reducing the number of jobs per build would have a larger impact on the congestion problem. For example, we could reduce the build matrix to 1x1x4 (1 DB, 1 requirement level, 4 python versions) until a PR actually passes on Travis, only then testing with a full build.
Or we could refrain from working directly on the WIP PRs and only update them when we _need_ to see if the build passes.

otron:
After re-examining this PR's builds, the time elapsed does suggest that it is an improvement:
- [Build #385](https://travis-ci.org/inveniosoftware/invenio-accounts/builds/102936839): 10m 1s
- [Build #386](https://travis-ci.org/inveniosoftware/invenio-accounts/builds/102939218): 20m 19s
- [Build #387](https://travis-ci.org/inveniosoftware/invenio-accounts/builds/102949869): 11m 7s
- [Build #388](https://travis-ci.org/inveniosoftware/invenio-accounts/builds/102950614): 11m 47s

While other PR builds seem to mostly be in the 14-18 minute range, with some spiking at 20-24 minutes but none lower than 14 minutes.

kaplun:
Hi! I wonder how this large build matrix are impacting other projects. Is Travis imposing a quota across organizations? Or users? Lately we are noticing large delays also in pushing inspirehep projects, and we were wondering if this is because the user triggering the build is also in many other projects and thus his quota is reduced in Travis.

jirikuncar:
@kaplun the important part is that we are using Open Source plan on Travis that encourages *"Fair use"* of resources. There is overall quota per repository depending on the system load (2-4 simultaneous tasks per repository). You can subscribe to https://www.traviscistatus.com/ to get notification when they have queue or other issues.

kaplun:
@jirikuncar Thanks. Indeed our issue then seems to corresponds to:
> [...]
> Jan 13, 14:44 UTC Build delays for travis-ci.org (public) sudo enabled Precise and Trusty builds
> [...]

So that explains it. 
Thanks for the pointer.

.travis.yml
```diff
@@ -68,13 +69,11 @@ install:
   - "travis_retry pip install -e .[${EXTRAS}]"
 
 before_script:
-  - "export DISPLAY=:99.0" #set up xvfb, req'd for non-phantomJS webdrivers
-  - "sh -e /etc/init.d/xvfb start"
-  - "export E2E_WEBDRIVER_BROWSERS='Firefox'"
-  - "export E2E_WEBDRIVER_TIMEOUT='300'" # Max allowed time for E2E tests
+  - "if [ ${E2E} == 'yes' ]; then export DISPLAY=:99.0; sh -e /etc/init.d/xvfb start; export E2E_WEBDRIVER_BROWSERS='Firefox'; export E2E_WEBDRIVER_TIMEOUT='300'; fi"
```
jirikuncar: `- "if [ ${E2E} == 'yes' ]; then e2e_setup.sh; fi"`?

otron: That was the original plan, but for some reason it didn't work on Travis.
I wanted to check if it'd work like this (which it did), then forgot to remove the `e2e_setup.sh` file, apparently.

jirikuncar: did you `chmod u+x e2e_setup.sh` in git?

otron: I did `+x` - the script was executed, but the E2E tests failed with the
same error you'd get if the commands had not been executed.

PR #94, opened 2016-01-20
tests: generate email for test users

### LINKED ISSUE #93: tests: repeated arg-less calls to `create_test_user` throw error
Calling `testutils.py:create_test_user` without any arguments twice cause an error to be thrown as it attempts to create a second user with the default email address.

Fix: generate random email address when none is given instead of using a static default.

otron:
* `testutils.py:create_test_user` generates a random email for test users when none is given, instead of
  using a static default (closes #93)

Signed-off-by: Odd Magnus Trondrud <odd.magnus.trondrud@cern.ch>

### DIFF
--- a/invenio_accounts/testutils.py
+++ b/invenio_accounts/testutils.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2015 CERN.
+# Copyright (C) 2015, 2016 CERN.
 #
 # Invenio is free software; you can redistribute it
 # and/or modify it under the terms of the GNU General Public License as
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -47,6 +47,7 @@
     'Flask-CeleryExt>=0.1.0',
     'Flask-CLI>=0.2.1',
     'Flask-Mail>=0.9.1',
+    'fake-factory>=0.5.3',
```
jirikuncar: Do we really need it?

invenio_accounts/testutils.py
```diff
@@ -35,6 +35,7 @@
 
 import flask
 import flask_login
+from faker import Faker
```
jirikuncar: IMHO it's not cool if we have dependency in package which is installed only for tests without checking it with `pkg_resources` at least.

otron: Does the `test_require` not declare dependencies needed for the tests but not installation?

setup.py
otron: No, we could do something like
```
import string
import random
def randstring(len=10, select_from=None):
    select_from = select_from or string.ascii_letters + '.' + string.digits
    return ''.join(random.SystemRandom().choice(select_from) for x in range(0, len))

email = randstring(10) + '@' + randstring(7) + '.net'
```

otron: ...Which should work just as well, I guess. I just figured using something existing would be less messy than writing it from scratch.

jirikuncar: I would prefer no automatic name generation - force developers to write tests with explicit values.

jirikuncar:
- [ ] add description to commit message and brackets around `(closes #93)`.

setup.py
otron: Why would you want to make writing tests more cumbersome?
Unless we're testing the validation part of user registration/creation, the actual values used are irrelevant.

invenio_accounts/testutils.py
jirikuncar: Yes, but `testutils.py` lives inside package. Hence one needs to be careful.

jirikuncar:
I still have general concern. If we use random string generation we might have occasional test failures, hence I oppose this change.

otron:
Only if the generator is capable of producing an invalid email address.

Is there some permutation of ascii letters and digits that is not a valid
email address prefix? (before the @)

However I agree that determinism is important for tests. We can provide a
static seed to the RNG, ensuring that the generated address is the same
between runs.

jirikuncar:
The generator can generate two same email addresses.

lnielsen:
Closing as discussed IRL

PR #100, opened 2016-01-26
config: allow for configuration of Redis session connection

### LINKED ISSUE #99: session backend needs to be configurable.
Set CONFIG parameter for session backend. Otherwise it tries to use localhost:6379 as a back end. In production we have nodes dedicated as a redis host so this needs to be a variable.

See
https://github.com/inveniosoftware/invenio-accounts/blob/master/invenio_accounts/ext.py#L67

We need something like.
```python
r = redis.StrictRedis(host=config['SESSION_REDIS_HOST'], port=config['SESSION_REDIS_PORT'], db=config['SESSION_REDIS_DB'])
```

We already have CACHE_REDIS_HOST as a config param. If we could use this by default, that would be great. The only annoying thing is that the port and db have to be specified, so the host can't be the full connection string.

eamonnmag:
* Adds 'ACCOUNTS_SESSION_REDIS_URL' as a configurable variable to set the redis connection parameters. If not present, the default localhost:6379/0 parameters as used. (closes #99)

Signed-off-by: Eamonn Maguire <eamonnmag@gmail.com>

### DIFF
--- a/invenio_accounts/config.py
+++ b/invenio_accounts/config.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2016 CERN.
+#
+# Invenio is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the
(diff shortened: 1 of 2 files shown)


invenio_accounts/config.py
```diff
@@ -0,0 +1 @@
+ACCOUNTS_SESSION_REDIS_URL = 'redis://localhost:6379/0'
```
jirikuncar: missing license

invenio_accounts/ext.py
```diff
@@ -63,8 +64,16 @@ def init_app(self, app, use_celery=True, sessionstore=None):
         if sessionstore is None:
             import redis
             from simplekv.memory.redisstore import RedisStore
-            # TODO: use default from config?
-            sessionstore = RedisStore(redis.StrictRedis())
+
+            # ACCOUNTS_SESSION_REDIS_URL should be in the form
+            # redis://username:password@host:port/db_index
+            if app.config['ACCOUNTS_SESSION_REDIS_URL']:
```
jirikuncar:
```python
redis_url = app.config['ACCOUNTS_SESSION_REDIS_URL']
if redis_url:
   sessionstore = RedisStore(redis.StrictRedis.from_url(redis_url))
```

PR #102, opened 2016-01-28
tests: E2E tests disabled locally by default

SamiHiltunen:
* Disables E2E tests locally by default. They can still be run by
  setting E2E_WEBDRIVER_BROWSERS.

Signed-off-by: Sami Hiltunen <sami.mikael.hiltunen@cern.ch>


jirikuncar:
FYI @lnielsen

PR #108, opened 2016-02-02
WIP sessions: nested transactions for create/delete

### LINKED ISSUE #98: sessions: create and delete should use nested transaction
- [ ] do not use Flask-Security datastore for managing SessionActivity
- [ ] verify that commit is called at after request when session is created/deleted

otron:
* Use `invenio_db.db` as interface to the database instead
  of flask-security's `datastore` (addresses #98)

* TODO: Nested transactions (closes #98)

Signed-off-by: Odd Magnus Trondrud <odd.magnus.trondrud@cern.ch>

### DIFF
--- a/invenio_accounts/sessions.py
+++ b/invenio_accounts/sessions.py
@@ -53,9 +53,11 @@ def add_session(session=None):
     """
     user_id, sid_s = session['user_id'], session.sid_s
     session_activity = SessionActivity(user_id=user_id, sid_s=sid_s)
-    with _db.session.begin_nested():
+    try:
+        with _db.session.begin_nested():
             _db.session.add(session_activity)
-    _db.session.commit()
+    except IntegrityError:
+        _db.session.rollback()
 
 
 def login_listener(app, user):
--- a/invenio_accounts/sessions.py
+++ b/invenio_accounts/sessions.py
@@ -53,11 +53,9 @@ def add_session(session=None):
     """
     user_id, sid_s = session['user_id'], session.sid_s
     session_activity = SessionActivity(user_id=user_id, sid_s=sid_s)
-    try:
-        _db.session.add(session_activity)
-        _db.session.commit()
-    except IntegrityError:
-        _db.session.rollback()
+    with _db.session.begin_nested():
+            _db.session.add(session_activity)
+    _db.session.commit()
 
 
 def login_listener(app, user):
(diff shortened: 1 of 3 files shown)


invenio_accounts/sessions.py
```diff
@@ -56,10 +54,10 @@ def add_session(session=None):
     user_id, sid_s = session['user_id'], session.sid_s
     session_activity = SessionActivity(user_id=user_id, sid_s=sid_s)
     try:
-        _datastore.db.session.add(session_activity)
-        _datastore.commit()
+        _db.session.add(session_activity)
```
jirikuncar: use `with db.session.begin_nested():` instead of try/except block

otron: Wait, we'd still need the try/except block otherwise an exception will just cause the entire app to crash, no?

jirikuncar: https://bitbucket.org/zzzeek/sqlalchemy/src/c8b3d4ed3f2638599fc73486cf0f724fa033a638/lib/sqlalchemy/orm/session.py?at=master&fileviewer=file-view-default#session.py-481:485

otron: So using `with begin_nested()` we're safe against exceptions thrown by `db.session.commit`. 
However exceptions thrown by statements inside the `with` block still need to be handled. Integrity Errors, for example, are thrown immediately if you try to add a duplicate entry to the session, not when you commit.

jirikuncar: `IntegrityError` happens during commit!

--
http://docs.sqlalchemy.org/en/latest/orm/session_state_management.html?highlight=integrityerror#merge-tips

otron: I'm getting one from `add`:
https://gist.github.com/otron/42d622a1a07ab0f5a444

invenio_accounts/sessions.py
```diff
@@ -34,14 +34,12 @@
 
 import flask
 from flask_login import user_logged_in
+from invenio_db import db as _db
```
jirikuncar: Why do you add the underscore prefix?

otron: For consistency with the other module-level variables.

otron:
@jirikuncar: Regarding [the discussion about try/except and begin_nested blocks](https://github.com/inveniosoftware/invenio-accounts/pull/108#discussion-diff-51550263) that's now on an "outdated diff":

- You were right about the `IntegrityError` being thrown by `db.session.commit()`, I misread the stack trace.
- The exit function of `begin_nested` does not catch the `IntegrityError` thrown when the `unique=True` constraint of `SessionActivity.sid_s` is violated. 

The following code snippet results in an unhandled `IntegrityError`:
```python
from invenio_db import db as _db
from invenio_accounts.models import SessionActivity

session_activity = SessionActivity(user_id=u'1', sid_s=u'stringval')
with _db.session.begin_nested():
    _db.session.add(session_activity)
    _db.session.add(session_activity)   
```

Placing the `with ...begin_nested()` block inside a `try` block catches the exception:
```python
from invenio_db import db as _db
from invenio_accounts.models import SessionActivity

session_activity = SessionActivity(user_id=u'1', sid_s=u'stringval')
try:
    with _db.session.begin_nested():
        _db.session.add(session_activity)
        _db.session.add(session_activity)
except IntegrityError:
    _db.session.rollback()
```

PR #116, opened 2016-03-09
examples: Menu initializion fix

jirikuncar:
* Removes menu initialization when InvenioTheme is enabled.
  (closes #10)

### DIFF
--- a/examples/app.py
+++ b/examples/app.py
@@ -126,13 +126,14 @@ FlaskCLI(app)
 Babel(app)
 Mail(app)
 InvenioDB(app)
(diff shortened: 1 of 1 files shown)


tiborsimko:
initializion -> initialisation

jirikuncar:
@tiborsimko I have fixed the typo. IMHO no announce is needed for example app.

lnielsen:
@tiborsimko So far we're not adding any FIX, NEW etc for the alpha releases, let's re-discuss if we want to change that.

PR #121, opened 2016-05-13
ext: separate REST extension

### LINKED ISSUE #120: global: different flask security blueprint handling for API and UI
Currently the flask-security blueprints are loaded in both API and UI.
After some discussions with @lnielsen and @jirikuncar there were two solutions:
1/ Load the blueprint only in one of the applications.
2/ Load them in both application but make the default content-type different: json for API and HTML for UI.

For the time being we chose to go with option 2. This remains as TODO but does not have a high priority right now.

jirikuncar:
* Separates REST extension that does not register blueprint.
  (closes #120)

--
cc @nharraud

### DIFF
--- a/invenio_accounts/config.py
+++ b/invenio_accounts/config.py
@@ -27,3 +27,7 @@
 # ACCOUNTS_SESSION_REDIS_URL should be in the form
 # redis://username:password@host:port/db_index
 ACCOUNTS_SESSION_REDIS_URL = 'redis://localhost:6379/0'
+
+# ACCOUNTS_FORCE_BLUEPRINT_REGISTRATION can be use to override
+# ``register_blueprint`` option.
+ACCOUNTS_REGISTER_BLUEPRINT = None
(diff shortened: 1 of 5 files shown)


nharraud:
@jirikuncar correct me if I am wrong, but without a config variable I don't see how one could enable `register_blueprint` when `invenio_base` is used to create the application.

jirikuncar:
@nharraud what about https://github.com/inveniosoftware/invenio-accounts/pull/121/commits/8811f08ac3bfb697a9f02250b79ad417ab6b5e7f ?

invenio_accounts/config.py
```diff
@@ -27,3 +27,7 @@
 # ACCOUNTS_SESSION_REDIS_URL should be in the form
 # redis://username:password@host:port/db_index
 ACCOUNTS_SESSION_REDIS_URL = 'redis://localhost:6379/0'
+
+# ACCOUNTS_FORCE_BLUEPRINT_REGISTRATION can be use to override
+# ``register_blueprint`` option.
+ACCOUTNS_REGISTER_BLUEPRINT = None
```
nharraud: s/ACCOUTNS_REGISTER_BLUEPRINT/ACCOUNTS_REGISTER_BLUEPRINT/g

nharraud:
@jirikuncar Apart from the little misspelling it looks good 👍  Thanks!

PR #123, opened 2016-05-18
global: support for legacy invenio password hashes

### LINKED ISSUE #122: global: support for legacy invenio password hashes
- Move code from:
  - https://github.com/inveniosoftware/invenio/blob/maint-2.1/invenio/ext/passlib/__init__.py
  - https://github.com/inveniosoftware/invenio/blob/maint-2.1/invenio/ext/passlib/__init__.py#L98
  - https://github.com/inveniosoftware/invenio/blob/maint-2.1/invenio/testsuite/test_ext_passlib.py

- Flask-Security: https://github.com/mattupstate/flask-security/blob/develop/flask_security/core.py#L93-L103
- Invenio-Accounts: https://github.com/inveniosoftware/invenio-accounts/blob/master/invenio_accounts/ext.py#L123-L124

omelkonian:
* NEW Adds the ability to migrate legacy Invenio users to Invenio 3
 automatically. The old encryption scheme has been integrated into the
 flask-security extension and a `migrate_hash` function can be used to
 update the password by re-encrypting it (using the current hash
 scheme) and storing it in the datastore. (closes #122)

 * Adds a new test utility, namely `create_legacy_user`.

 * Increases test coverage.

Signed-off-by: Orestis Melkonian <melkon.or@gmail.com>
Co-Authored-By: Lars Holm Nielsen <lars.holm.nielsen@cern.ch>

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -71,6 +71,7 @@ Invenio user management and authentication.
 - Mikael Karlsson <i8myshoes@gmail.com>
 - Nicholas Robinson <nicholas.robinson@cern.ch>
 - Nikolaos Kasioumis <nikolaos.kasioumis@cern.ch>
+- Orestis Melkonian <melkon.or@gmail.com>
 - Patrick Glauner <patrick.oliver.glauner@cern.ch>
 - Paulo Cabral <paulo.cabral@cern.ch>
 - Peter Halliday <phalliday@cornell.edu>
--- a/invenio_accounts/hash.py
+++ b/invenio_accounts/hash.py
@@ -1,0 +1,136 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2015, 2016 CERN.
+#
+# Invenio is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Legacy Invenio hash support."""
+
+import hashlib
+
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
+from passlib.utils.compat import str_to_uascii
+from passlib.utils.handlers import GenericHandler, HasSalt, parse_mc2, \
+    render_mc2
+from six import PY3, binary_type, text_type
+
+__all__ = ('invenio_aes_encrypted_email', )
+
+
+def _to_binary(val):
+    """Convert to binary."""
+    if isinstance(val, text_type):
+        return val.encode('utf-8')
+    assert isinstance(val, binary_type)
+    return val
+
+
+def _to_string(val):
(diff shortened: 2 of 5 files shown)


setup.py
```diff
@@ -92,6 +92,7 @@
     'Flask-KVSession>=0.6.1',
     'SQLAlchemy-Utils[ipaddress]>=0.31.0',
     'redis>=2.10.5',
+    'cryptography>=0.6',
```
jirikuncar: can you keep the requirements sorted?

AUTHORS.rst
```diff
@@ -83,3 +83,4 @@ Invenio user management and authentication.
 - Tibor Simko <tibor.simko@cern.ch>
 - Trond Aksel Myklebust <trond.aksel.myklebust@cern.ch>
 - Yoan Blanc <yoan.blanc@cern.ch>
+- Orestis Melkonian <melkon.or@gmail.com>
```
lnielsen: The list is ordered alphabetical by first name

invenio_accounts/testutils.py
```diff
@@ -66,6 +66,27 @@ def create_test_user(email='test@test.org',
     return user
 
 
+def create_legacy_user(email='test@test.org', password='123456', **kwargs):
```
lnielsen: I don't think we should add this method. The ``create_test_user`` is an API function used by other modules to create a user in a fast and easy way, and we don't want to allow other modules to create users with legacy passwords.

lnielsen: I.e. you can move this method to the pytest helpers since this file is for code that other modules can import.

tests/test_utils.py
```diff
@@ -144,3 +144,35 @@ def test_login_user_via_view(app):
             testutils.login_user_via_view(client, user.email,
                                           user.password_plaintext)
             assert testutils.client_authenticated(client)
+
+
+def test_create_legacy_user(app):
+    """Test for testutils.py:create_legacy_user().
+
+        Demonstrates basic usage and context requirements.
```
lnielsen: indention

invenio_accounts/hash.py
```diff
@@ -0,0 +1,142 @@
...
+    name = "invenio_aes_encrypted_email"
+
+    def _calc_checksum(self, secret):
+        """Calculate string."""
+        return str_to_uascii(
+            hashlib.sha256(mysql_aes_encrypt(self.user, secret)).hexdigest()
+        )
+
+
+def migrate_hash(user):
```
lnielsen: Is this method needed? I think Flask-Security will do this automatically when a user login.

lnielsen:
Looks like a good start 👍 

- A high level test case where a user has a Invenio legacy hash, then login the user via ``/login`` and verify that 1) user is authenticated, and 2) the password hash has been upgraded.

E.g. I would expect that the password hash would need to be in the format ``$algoid$salt_ie_email$hash_value`` in order for passlib to which algo to use.

invenio_accounts/hash.py
omelkonian: Well, there is a problem with the `user` context keyword that the legacy scheme uses. `flask-security`'s built-in methods `verify_password` and `encrypt_password` do not accept context arguments, therefore there is no way to not interfere with the internal `_pwd_context`. Maybe I am missing something...

omelkonian:
OK, will look into it.

invenio_accounts/hash.py
lnielsen: This is related to my comment below about the hash format:  

> E.g. I would expect that the password hash would need to be in the format $algoid$salt_ie_email$hash_value in order for passlib to know which algo to use.

tox.ini
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
```
lnielsen: We're normally not using Tox anymore. Usually Tox is not used enough by developers and thus the tox.ini ends up not being maintained properly.

lnielsen:
LGTM

tox.ini
omelkonian: But how do you test for both python 2 and 3? Do you use two distinct virtualenvs, one for each version?

lnielsen: I normally let Travis do the job. And in case I detect problems related to python version, i'll just create two virtualenvs. It's rare that I continuesly want to run all tests on both python versions because of the time it takes

lnielsen:
I'll merge as soon as Travis is green

PR #128, opened 2016-06-03
views: allow CHANGEABLE to be False

omelkonian:
* FIX Fixes issue of `change_password` registration, even when the
   configuration `variable SECURITY_CHANGEABLE` was set to `False`.


jirikuncar:
@omelkonian next time you can just force push to the same branch when you need to update a PR (`git push origin bugfix-changeable --force`) instead creating #129.

PR #129, opened 2016-06-03
views: allow CHANGEABLE to be False

omelkonian:
* FIX Fixes issue of `change_password` registration, even when the
   configuration `variable SECURITY_CHANGEABLE` was set to `False`.

Signed-off-by: Orestis Melkonian <melkon.or@gmail.com>

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2015 CERN.
+# Copyright (C) 2015, 2016 CERN.
 #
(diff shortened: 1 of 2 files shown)


egabancho:
Travis is failing because of Sphinx, could you please take a look at https://github.com/inveniosoftware/cookiecutter-invenio-module/commit/c8d93e4f06be8f329bb25ea69c19a53ffbcc980a?

omelkonian:
Working

egabancho:
Maybe is better to split your changes into two commits, one for the changeable thing and the other for Sphinx. It makes the commit history better and might help in the future with `git bisect` and friends

PR #155, opened 2016-09-09
i18n: translation improvements

alizeepace:
* Corrects Czech translations as suggested by @helix84.

* Adds German translations.


alizeepace:
@tiborsimko I think a fluent German speaker should check my translations :smile:

PR #156, opened 2016-09-12
config: turn on secure session cookie flag

liamkirsh:
Since Invenio should be running over SSL, this option sets the secure flag on the session cookie to avoid session stealing through a man-in-the-middle attack. (I believe this will break login over insecure connections, but @kaplun suggested that we can just set up SSL on INSPIRE Nightly.)

Signed-off-by: Liam Kirsh <liam.kirsh@cern.ch>

### DIFF
--- a/invenio_accounts/views.py
+++ b/invenio_accounts/views.py
@@ -91,3 +91,15 @@ def init_menu():
             _("Change password"),
             order=0,
         )
+
+
+@blueprint.before_app_first_request
+def check_security_settings():
+    """Warn if session cookie is not secure in production."""
+    in_production = not (current_app.debug or current_app.testing)
+    secure = current_app.config.get('SESSION_COOKIE_SECURE')
+    if in_production and not secure:
+        current_app.logger.warning(
+            "SESSION_COOKIE_SECURE setting must be set to True to prevent the "
+            "session cookie from being leaked over an insecure channel."
+        )


invenio_accounts/config.py
```diff
@@ -31,3 +31,7 @@
 # ACCOUNTS_FORCE_BLUEPRINT_REGISTRATION can be use to override
 # ``register_blueprint`` option.
 ACCOUNTS_REGISTER_BLUEPRINT = None
+
+# Set the secure flag on the session cookie to ensure it is only transmitted
+# over a secure channel.
+SESSION_COOKIE_SECURE = True
```
lnielsen: We have to make it work for using the development server as well. Like this, the dev server will be unusable as it's running http.....

jirikuncar: @liamkirsh this default value will not be loaded because of (a) https://github.com/inveniosoftware/invenio-accounts/blob/master/invenio_accounts/ext.py#L202-L204 it doesn't have `ACCOUNTS_` prefix (b) Flask already sets default value for this option.

jirikuncar: What about just creating a warning before first request if the `SESSION_COOKIE_SECURE` is `False` and we are not in debug or testing mode?

liamkirsh:
Would views.py be the proper place to put that?

invenio_accounts/views.py
```diff
@@ -91,3 +91,14 @@ def init_menu():
             _("Change password"),
             order=0,
         )
+
+
+@blueprint.before_app_first_request
+def check_security_settings():
+    """Warn if session cookie is not secure in production."""
+    if (not current_app.debug and not current_app.testing and
```
jirikuncar: I was trying to understand the condition and make it more readable. What do you think about following code?

```python
in_production = not (current_app.debug or current_app.testing)
secure = current_app.config.get('SESSION_COOKIE_SECURE')
if in_production and not secure:
    current_app.logger.warning(...)
```

jirikuncar:
@liamkirsh please keep the PR up-to-date with the master branch by rebasing. Thanks!

invenio_accounts/views.py
liamkirsh: I agree, thanks.

Issue #154, opened 2016-09-08
admin: user creation doesn't work


PXke:
If you use different way to create user. 

With the /signup/ url, or the admin interface, or the cli. It seems they all ignore each others. 

For example if I create 4 users with admin it will allocate 1-4 for ids then if I use the CLI I will have to crash it four times to make it create a user with the id 5. 

Way to create, ID 
Admin UI, 1
Admin UI, 2
Admin UI, 3
CLI, 1 -> crash 
CLI, 2 -> crash
CLI, 3 -> crash
CLI, 4

lnielsen:
This seems to be a problem with the admin interface - honestly I think user creation via admin haven't been tested at all. For once, the admin shouldn't ask to fill in the id, also it shouldn't allow editing/changing some of the timestamps.

<img width="745" alt="screen shot 2016-09-08 at 14 42 59" src="https://cloud.githubusercontent.com/assets/1698163/18349628/8f297cae-75d2-11e6-972f-2eec9a7e26ca.png">

lnielsen:
Probably the creation form should only have:

- Email
- Active

We could enable "confirmed at", which controls if an account's email address was confirmed or not so that an admin can confirm an email address for a given user, but not sure it's a good idea.

PR #157, opened 2016-09-15
accounts: create/edit user form fix

diegodelemos:
* Makes available just email and active fields.

Signed-off-by: Diego Rodriguez <diego.rodriguez.rodriguez@cern.ch>

### DIFF
--- a/tests/test_admin.py
+++ b/tests/test_admin.py
@@ -25,13 +25,12 @@
 import pytest
 from flask import current_app, url_for
 from flask.ext.security.utils import encrypt_password
-from flask_admin import Admin, menu
+from flask_admin import menu
 from invenio_admin import InvenioAdmin
 from invenio_db import db
 from werkzeug.local import LocalProxy
 
 from invenio_accounts import InvenioAccounts
-from invenio_accounts.admin import role_adminview, user_adminview
 from invenio_accounts.cli import users_create
 
 _datastore = LocalProxy(
@@ -39,21 +38,8 @@ _datastore = LocalProxy(
 )
 
 
-def test_admin(app):
+def test_admin(app, admin_view):
     """Test flask-admin interace."""
-    assert isinstance(role_adminview, dict)
-    assert isinstance(user_adminview, dict)
-
-    assert 'model' in role_adminview
-    assert 'modelview' in role_adminview
-    assert 'model' in user_adminview
-    assert 'modelview' in user_adminview
-
-    admin = Admin(app, name="Test")
-
-    user_model = user_adminview.pop('model')
-    user_view = user_adminview.pop('modelview')
-    admin.add_view(user_view(user_model, db.session, **user_adminview))
 
     # Test activation and deactivation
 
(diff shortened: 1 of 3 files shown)


tests/test_admin.py
```diff
@@ -28,6 +28,7 @@
 from flask_admin import Admin, menu
 from invenio_admin import InvenioAdmin
 from invenio_db import db
+from six import b
```
jirikuncar: do you need this import?

tests/test_admin.py
```diff
@@ -92,3 +93,27 @@ def test_admin(app):
             data={'rowid': -42, 'action': 'activate'},
             follow_redirects=True
         )
+
+        # Test empty mail form
+
+        res = client.post(
+            url_for("user.create_view"),
```
jirikuncar: Please use preferably single quotes if possible. Thanks

tests/test_admin.py
```diff
@@ -92,3 +93,27 @@ def test_admin(app):
...
+            data={'email': ''},
+            follow_redirects=True
+        )
+        assert b'This field is required.' in res.data
+
+        # Test user creation
+
+        res = client.post(
+            url_for("user.create_view"),
```
jirikuncar: `s/"/'/`

tests/test_admin.py
```diff
@@ -92,3 +93,27 @@ def test_admin(app):
...
+            data={'email': 'test2@test.cern'},
+            follow_redirects=True
+        )
+        assert _datastore.get_user('test2@test.cern') is not None
+
+        user_data = dict(email="test3@test.cern", active=False,
```
jirikuncar: `s/"/'/`

tests/test_admin.py
```diff
@@ -92,3 +93,27 @@ def test_admin(app):
...
+                         password=encrypt_password('aafaf4as5fa'))
```
jirikuncar: Simple possible password is preferable in tests: `123456`

tests/test_admin.py
```diff
@@ -92,3 +93,27 @@ def test_admin(app):
...
+        _datastore.create_user(**user_data)
```
jirikuncar: IMHO this should be part of new test case that simulates flow defined in https://github.com/inveniosoftware/invenio-accounts/issues/154

tests/test_admin.py
diegodelemos: Yes, I need it to make this line compatible with Python 3:
`assert b'This field is required.' in res.data`

jirikuncar: `b'foo'` works on Python 2.7.

tests/test_admin.py
```diff
@@ -92,3 +93,69 @@ def test_admin(app):
             data={'rowid': -42, 'action': 'activate'},
             follow_redirects=True
         )
+
+    user_adminview['model'] = user_model
```
jirikuncar: why do you need it?

jirikuncar: Maybe line 55 and 56 should use `get` instead of `pop`. (`user_model = user_adminview.get('model')`)

diegodelemos: Fixed copying `user_adminview` using `dict` constructor

tests/test_admin.py
```diff
@@ -92,3 +93,67 @@ def test_admin(app):
...
+        res = client.post(
+            url_for('user.create_view'),
+            data={'email': 'test1@test.cern'},
+            follow_redirects=True
+        )
+        assert _datastore.get_user('test1@test.cern') is not None
+
+        res = client.post(
+            url_for('user.create_view'),
+            data={'email': 'test2@test.cern'},
```
jirikuncar: check also `active=True` and `active=False` for test2 and test3 respectively

tests/test_admin.py
```diff
@@ -92,3 +93,67 @@ def test_admin(app):
             data={'rowid': -42, 'action': 'activate'},
             follow_redirects=True
         )
+
+
+def test_admin_createuser(app):
+    """Test flask-admin user creation"""
+    assert isinstance(user_adminview, dict)
```
jirikuncar: consider creating `admin` fixture in `conftest.py` from lines 100-110 (and 44-56)

PR #177, opened 2017-02-03
global: fix release checklist

remileduc:
* FIX Adds changes according to release preparation checklist.

Signed-off-by: Rémi Ducceschi <remi.ducceschi@gmail.com>

*NOTE*: the USAGE documentation should be updated to give a real use case. Currently, it is the same as the API documentation...

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2015 CERN.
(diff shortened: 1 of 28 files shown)


remileduc:
the USAGE documentation should be updated to give a real use case. Currently, it is the same as the API documentation...

remileduc:
some tests doesn't pass, need to investigate...

hachreak:
apart some small things, the rest is fine I think! :+1:

PR #178, opened 2017-02-21
tasks: new task to clean up expired sessions

hachreak:
* Adds a new task to clean up expired sessions. (addresses #52)

### DIFF
--- a/invenio_accounts/tasks.py
+++ b/invenio_accounts/tasks.py
@@ -29,6 +29,11 @@ from __future__ import absolute_import, print_function
 from celery import shared_task
 from flask import current_app
 from flask_mail import Message
+from invenio_db import db
+
+from .models import SessionActivity
+from .proxies import current_accounts
+from .sessions import delete_session
 
 
 @shared_task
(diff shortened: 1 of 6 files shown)


invenio_accounts/tasks.py
```diff
@@ -40,3 +45,12 @@ def send_security_email(data):
     msg = Message()
     msg.__dict__.update(data)
     current_app.extensions['mail'].send(msg)
+
+
+@shared_task
+def clean_session_table():
+    """Clean session table."""
+    expired_moment = datetime.now() - current_app.permanent_session_lifetime
```
lnielsen: i think this should be ``datetime.utcnow()``

invenio_accounts/tasks.py
```diff
@@ -40,3 +45,12 @@ def send_security_email(data):
...
+    SessionActivity.query.filter(
```
lnielsen: Perhaps we could add it as a method on SessionActivity...something like:

``SessionActivity.query_by_expired().delete()``

hachreak:
Task documentation added!

hachreak:
ping  @lnielsen

PR #179, opened 2017-02-22
views: manage security table for the user 

hachreak:
* Adds a view to manage the security session table. (addresses #52)

### DIFF
--- a/invenio_accounts/views.py
+++ b/invenio_accounts/views.py
@@ -26,13 +26,22 @@
 
 from __future__ import absolute_import, print_function
 
-from flask import Blueprint, current_app
+from flask import Blueprint, current_app, flash, redirect, render_template, \
+    request, url_for
 from flask_babelex import lazy_gettext as _
-from flask_menu import current_menu
+from flask_breadcrumbs import register_breadcrumb
+from flask_login import login_required
+from flask_menu import current_menu, register_menu
+from flask_security import current_user
+from invenio_db import db
+
+from .models import SessionActivity
+from .sessions import delete_session
 
 blueprint = Blueprint(
     'invenio_accounts',
     __name__,
+    url_prefix='/account/settings',
     template_folder='templates',
     static_folder='static',
 )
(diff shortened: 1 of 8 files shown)


hachreak:
(rebased on top of #178 )

invenio_accounts/views.py
```diff
@@ -103,3 +112,39 @@ def check_security_settings():
...
+@login_required
+@register_menu(
+    blueprint, 'settings.security',
+    # NOTE: Menu item text (icon replaced by a user icon).
+    _('%(icon)s Security', icon='<i class="fa fa-user fa-fw"></i>'),
+    order=0)
+@register_breadcrumb(blueprint, 'breadcrumbs.settings.security', _('Security'))
+def security():
+    """View for security page."""
+    sessions = SessionActivity.query.filter_by(user_id=current_user.id).all()
```
lnielsen: I would add query methods to SessionActivity since it's likely we need might need them other places:

```SessionActivitiy.query_by_user(current_user.get_id()).all()```

You could also add an instance method:

```session_activity.is_current_session() -> session.id == self.sid_s```. Then it's easy from the template to figure out if a session is the current one.

I think you need to use ``current_user.get_id()`` instead of ``current_user.id`` (i.e. check API in Flask-Login).

File: invenio_accounts/views.py [hunk 2]
```diff
@@ -103,3 +112,39 @@ def check_security_settings():
...
+    return render_template(
+        current_app.config['ACCOUNTS_SETTINGS_SECURITY_TEMPLATE'],
+        sessions=sessions
+    )
+
+
+# TODO use a DELETE?
+@blueprint.route('/revoke_session', methods=['GET'])
```
lnielsen: I would go with ``POST``.

invenio_accounts/views.py
```diff
@@ -103,3 +112,39 @@ def check_security_settings():
...
+@login_required
+def revoke_session():
+    """Revoke a session."""
+    sid_s = request.args['sid_s']
+    if SessionActivity.query.filter_by(
```
lnielsen:
```
request.form['sid_s'] # to ensure you get it from POST body
```

setup.cfg
```diff
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+
```
lnielsen: Not needed i think :-)

lnielsen:
Can you also add a small documentation note in the usage guide for how to put the Celery task in the cron schedule.... (note, we should actually add documentation to celery module how to run cron tasks - https://github.com/inveniosoftware/invenio-celery/issues/43). So documentation should just quickly mention what a user should do to enable the celery task

hachreak:
PR updated!

hachreak:
ping @lnielsen

tests/test_views.py
```diff
@@ -24,11 +24,13 @@
 
 """Test account views."""
 
+import mock
```
tiborsimko: `mock` is missing in `setup.py`, please add it to `tests_require` there.

hachreak:
ping @lnielsen  @tiborsimko

PR #180, opened 2017-03-01
admin: session administration addition, session creation on sign-up

### LINKED ISSUE #190: global: session activity not created on sign-up
**Reproduce**

1. Setup and load fixtures for example app.  
1. Run example app
1. Go to http://localhost:5000/signup/ and sign up. (after sign up you will be automatically logged in)
1. Inspect database and see that no session was created:

```console
$ sqlite3 examples/instance/app.db
SQLite version 3.14.0 2016-07-26 15:17:14
Enter ".help" for usage hints.
sqlite> select count(*) from accounts_user_session_activity;
0
```

To see that it works with login:

1. Setup and load fixtures for example app.  
1. Run example app
1. Go to http://localhost:5000/login/ and login with info@inveniosoftware.org / 123456
1. Inspect database and see that no session was created:

```console
$ sqlite3 examples/instance/app.db
SQLite version 3.14.0 2016-07-26 15:17:14
Enter ".help" for usage hints.
sqlite> select count(*) from accounts_user_session_activity;
1
```

hachreak:
* Adds a session administration page. (addresses #52)

* Commits the session when login permit to fix the session creation
  on user first signup. (closes #190)

### DIFF
--- a/invenio_accounts/admin.py
+++ b/invenio_accounts/admin.py
@@ -26,6 +26,7 @@
 
 from flask import current_app, flash
 from flask_admin.actions import action
+from flask_admin.babel import lazy_gettext
 from flask_admin.contrib.sqla import ModelView
 from flask_admin.contrib.sqla.ajax import QueryAjaxModelLoader
 from flask_admin.form.fields import DateTimeField
@@ -33,13 +34,15 @@ from flask_admin.model.fields import AjaxSelectMultipleField
 from flask_babelex import gettext as _
 from flask_security.recoverable import send_reset_password_instructions
 from flask_security.utils import encrypt_password
+from invenio_db import db
 from passlib import pwd
 from werkzeug.local import LocalProxy
 from wtforms.fields import BooleanField
 from wtforms.validators import DataRequired
 
 from .cli import commit
-from .models import Role, User
+from .models import Role, SessionActivity, User
+from .sessions import delete_session
 
 _datastore = LocalProxy(lambda: current_app.extensions['security'].datastore)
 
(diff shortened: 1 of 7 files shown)


lnielsen:
@hachreak Can you rebase on top of latest merged changes.
@switowski Will you review?

invenio_accounts/admin.py
```diff
@@ -137,6 +140,56 @@ class RoleView(ModelView):
...
+
+    form_ajax_refs = {
+        'user': {
+            'fields': (User.id, User.email)
+        },
+    }
+
+    def delete_model(self, model):
+        is_current = SessionActivity.is_current(sid_s=model.sid_s)
+        if is_current:
```
switowski: Just for the future reference, I would replace it with just:
`if SessionActivity.is_current(sid_s=model.sid_s)`
as you don't use this `is_current` variable anywhere else, so there is no benefit of assigning to this variable.
But that's just for the future, no need to change the PR because of this silly reason :)

hachreak: thanks... I forgot to remove it :smile:

hachreak:
I'll rebase it now and fix the conflict! :smile:

setup.py
```diff
@@ -55,7 +55,7 @@
         'celery>=3.1.0,<4.0',
     ],
     'docs': [
-        'Sphinx>=1.4.2',
+        'Sphinx>=1.4.2,<1.6',
```
lnielsen: Do we now have Sphinx related builds problems again?

hachreak: See https://github.com/inveniosoftware/troubleshooting/issues/11

PR #182, opened 2017-03-03
views: view sessions templates and CSRF fix

lnielsen:
* Improves design of templates a bit.

* Fixes a CSRF issue with revoke form.

Signed-off-by: Lars Holm Nielsen <lars.holm.nielsen@cern.ch>

### DIFF
--- a/invenio_accounts/templates/invenio_accounts/settings/security.html
+++ b/invenio_accounts/templates/invenio_accounts/settings/security.html
@@ -24,40 +24,42 @@
 #}
 {%- extends config.ACCOUNTS_SETTINGS_TEMPLATE %}
 
-{% set panel_title = _('Security') %}
-{% set panel_subtitle = _('This is a list of devices that have logged into your account.') %}
+{% set panel_title = _('Sessions') %}
 {% set panel_icon = 'fa fa-shield fa-fw' %}
+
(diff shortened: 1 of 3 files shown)


invenio_accounts/templates/invenio_accounts/settings/security.html
```diff
@@ -24,40 +24,42 @@
...
+  {%- set form = formclass(sid_s=session.sid_s) %}
   <div class="row record-elem">
     <div class="col-xs-1">
-      {% if is_current(session.sid_s) %}
-        <i class="fa fa-circle" aria-hidden="true"></i>
-      {% else %}
-        <i class="fa fa-circle-o" aria-hidden="true"></i>
-      {% endif %}
+      {%- if is_current_sid %}
+        <i class="fa fa-circle" style="color: #28a745;" aria-hidden="true"></i>
```
jirikuncar: What about using class `text-success` instead direct style color?

lnielsen: ``text-success`` is too dark green.

jirikuncar: Then it would be nice to have a new css class so one can simply override the css instead of whole template.

PR #187, opened 2017-03-09
installation: environment marker fix

lnielsen:
* Fixes issue with environment markers supported by setuptools and pip
  not being handled properly by wheel.

Signed-off-by: Lars Holm Nielsen <lars.holm.nielsen@cern.ch>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,6 +48,9 @@ tests_require = [
 ]
 
 extras_require = {
+    ':python_version=="2.7"': [
+        'ipaddr>=2.1.11',
+    ],
     'celery': [
         'celery>=3.1.0,<4.0',
     ],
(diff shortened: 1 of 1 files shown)


lnielsen:
Without this fix, the released wheels doesn't install on Python 3. With the fix in inveniosoftware/requirements-builder#27 it now probably handles conditions in extra_requires.

After digging into wheel source code I found that wheel simply strips of the env marker when it is in install requires. [This function](https://bitbucket.org/pypa/wheel/src/54ddbcc9cec25e1f4d111a142b8bfaa163130a61/wheel/metadata.py?at=default&fileviewer=file-view-default#metadata.py-230) strips it from install_requires. [This functions](https://bitbucket.org/pypa/wheel/src/54ddbcc9cec25e1f4d111a142b8bfaa163130a61/wheel/metadata.py?at=default&fileviewer=file-view-default#metadata.py-254) handles the condition in extra require.

PR #189, opened 2017-03-10
templates: empty session check in security.html

dinosk:
* Fixes 500 error when visiting security page
  immediately after signing up

Signed-off-by: Dinos Kousidis <konstantinos.kousidis@cern.ch>


lnielsen:
I looked a bit more into the problem, and actually it's somewhat more complicated than initial thought.

First problem is that ``session`` variable name is already a [Flask global variable](http://flask.pocoo.org/docs/0.12/templating/#standard-context). Hence we should fix the loop:

```jinja
{% for session in sessions %} -> {% for s in sessions %}
```

The second and real problem here is that when you sign up, the session is not saved to the database ``accounts_user_session_activity`` table.

lnielsen:
Closes PR as the problem is caused by #190

PR #193, opened 2017-03-17
config: unified url prefix for password change

jirikuncar:
See also url prefix in https://github.com/inveniosoftware/invenio-accounts/blob/master/invenio_accounts/views.py#L45


lnielsen:
Thanks

Issue #136, opened 2016-07-04
accounts: delete account feature


jirikuncar:
@lnielsen commented on [Mon Aug 04 2014](https://github.com/inveniosoftware/invenio/issues/2023)

**Goal:**
Allow users to delete their account via the account settings.

**Additional info**
The account deletion facility should check any attached baskets, any predefined alerts, any submitted records, messages etc.

Overlays must be able to hook into the deletion process, to do custom processing via e.g. pre/post delete signals.

**Open questions:**
Do we delete the account or simply mark it as deleted for record keeping? 

See also issue #1973 

---

@jirikuncar commented on [Mon Aug 04 2014](https://github.com/inveniosoftware/invenio/issues/2023#issuecomment-51028833)

> Overlays must be able to hook into the deletion process, to do custom processing via e.g. pre/post delete signals.

I think we can rely on SQLAlchemy signals for `User` model.

---

@tiborsimko commented on [Wed Aug 13 2014](https://github.com/inveniosoftware/invenio/issues/2023#issuecomment-52047848)

> Do we delete the account or simply mark it as deleted for record keeping? 

I think the account inactivation and marking-as-deleted would be good for record keeping.  We do it already for records in the same way.  (via `980 $c DELETED`)

---

@aw-bib commented on [Wed Aug 13 2014](https://github.com/inveniosoftware/invenio/issues/2023#issuecomment-52051190)

I think one point needs to be addressed here which shares probably some parts with #2022.

Today, the more common case with email addresses in research institutions and universities is, that the email address as such get's recycled. So you may have a naming scheme like initial.lastname@institution.somewhere (say j.doe@example.com). Now if you have only one j.doe@example.com he gets this address. Now, j.doe@example.com may leave the company. Later a another one joins in some John Doe, and he gets j.doe@example.com...

To make the long story short: IRL email addresses do not make a good primary key anymore :(

---

@tiborsimko commented on [Wed Aug 13 2014](https://github.com/inveniosoftware/invenio/issues/2023#issuecomment-52056040)

@aw-bib Yes, e.g. Yahoo email addresses may also get recycled now... http://yahoo.tumblr.com/post/52805929240/yourname-yahoo-com-can-be-yours

PR #199, opened 2017-05-10
Admin changes

jmartinm:


invenio_accounts/admin.py
```diff
@@ -129,13 +134,14 @@ class RoleView(ModelView):
     list_all = ('id', 'name', 'description')
 
     column_list = \
-        form_columns = \
         column_searchable_list = \
         column_filters = \
```
jirikuncar: do we really want to allow admins to search by password?

lnielsen:
Thanks for the PR. Some quick questions at first sight:

1) Have you checked the UI with large number of users 1000+? I've usually had problems with admin ui when allowing editing related objects when there's a large number.
2) Deleting users is a bit tricky. The admin view shouldn't care much, but there's quite a number of things that can fail if you delete a user. See a bit of previous discussion in #136. The easy fix is to make it configurable right now with the default to off.

invenio_accounts/admin.py
jmartinm: The password is only added to `form_columns` so that it is editable but it will not be displayed in the list view or be searchable.

jmartinm:
> Have you checked the UI with large number of users 1000+? I've usually had problems with admin ui when allowing editing related objects when there's a large number.

Just gave it a try with 10000 users and indeed the users field becomes slow (it's usable, but barely)

screen shot 2017-05-11 at 10 49 56 am

We might need some custom widget that does not load all the users on the browser but autocompletes from the server side.

jmartinm:
Updated the PR so that `users` are loaded via AJAX, now it works fast.

ajax-request

jmartinm:
> Deleting users is a bit tricky. The admin view shouldn't care much, but there's quite a number of things that can fail if you delete a user. See a bit of previous discussion in #136. The easy fix is to make it configurable right now with the default to off.

I deleted the commit about deleting users for the moment, since indeed I am not sure of what the consequences can be ;)

jmartinm:
I have update the PR based on the conversation we had yesterday @lnielsen 

The workflow is the following (inspired by the workflow that WordPress uses):

1) When creating a new user from the admin. A default 'strong' password is set for you:

screen shot 2017-05-12 at 2 05 09 pm

2) A checkbox is present `Send User Notification`. When marked, the new user will receive an email to reset their password (using [flask-security](https://github.com/mattupstate/flask-security/blob/d76198d078af3cba559085f4bba8fbdcbb96c2e0/flask_security/recoverable.py#L25))

invenio_accounts/admin.py
```diff
@@ -75,6 +88,14 @@ class UserView(ModelView):
         'last_login_ip': _('Last Login IP')
     }
 
+    def on_model_change(self, form, User, is_created):
+        if form.password.data is not None:
+            User.password = encrypt_password(form.password.data)
+
+    def after_model_change(self, form, User, is_created):
+        if is_created and form.notification.data is True:
```
jirikuncar: `form.notification.data is True` or just `form.notification.data`?

invenio_accounts/utils.py
```diff
@@ -0,0 +1,37 @@
...
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Utils for Invenio Accounts."""
+
+import random
+
+
+def generate_password(length=12, special_chars=True, extra_special_chars=False):
```
jirikuncar: consider using https://passlib.readthedocs.io/en/stable/lib/passlib.utils.html#passlib.utils.generate_password

PR #203, opened 2017-05-16
global: support disabling session activity tab

### LINKED ISSUE #201: global: support disabling session activity tab

hachreak:
* Supports disabling session activity tab through configuration.
  (closes #201)

### DIFF
--- a/examples/app.py
+++ b/examples/app.py
@@ -25,8 +25,6 @@
 
 u"""Minimal Flask application example for development.
 
-Run Redis server.
-
 Install the Invenio default theme
 
 You should execute these commands in the examples-directory.
@@ -73,7 +71,7 @@ from invenio_db import InvenioDB
 from invenio_i18n import InvenioI18N
 
 from invenio_accounts import InvenioAccounts
-from invenio_accounts.views import blueprint
+from invenio_accounts.views.settings import blueprint
 
 try:
     pkg_resources.get_distribution('invenio_assets')
--- a/invenio_accounts/config.py
+++ b/invenio_accounts/config.py
@@ -50,6 +50,12 @@ ACCOUNTS_USE_CELERY = None
 By default, it uses Celery if it can find it.
 """
 
+ACCOUNTS_SESSION_ACTIVITY_ENABLED = True
+"""Enable session activity store on database."""
+
+ACCOUNTS_SETTINGS_SECURITY_TEMPLATE = 'invenio_accounts/settings/security.html'
+"""Template for the account security page."""
+
 # Change Flask-Security defaults
 SECURITY_CHANGEABLE = True
 """
@@ -209,5 +215,3 @@ SECURITY_CHANGE_URL = '/account/settings/password/'
    <https://pythonhosted.org/Flask-Security/configuration.html>`_
    configuration.
 """
-
-ACCOUNTS_SETTINGS_SECURITY_TEMPLATE = 'invenio_accounts/settings/security.html'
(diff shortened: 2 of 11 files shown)


hachreak:
waiting for #180

invenio_accounts/ext.py
```diff
@@ -192,6 +192,15 @@ def init_config(self, app):
             if any([k.startswith(prefix) for prefix in config_apps]):
                 app.config.setdefault(k, getattr(config, k))
 
+    def _enable_session_activity(self, app):
+        """Enable session activity."""
+        user_logged_in.connect(login_listener, app)
+        user_logged_out.connect(logout_listener, app)
+        from .views.settings import blueprint
```
lnielsen: I think normally imports inside functions should not use relative imports.

examples/app.py
```diff
@@ -25,6 +25,8 @@
 
 u"""Minimal Flask application example for development.
 
+Run Redis server.
```
switowski: There is already mention about the redis later in the documentation, so maybe we shouldn't duplicate it: https://github.com/hachreak/invenio-accounts/blob/b2620e66978ebad95da50f01f430efaffdfe793f/examples/app.py#L42

invenio_accounts/config.py
```diff
@@ -50,6 +50,12 @@
 By default, it uses Celery if it can find it.
 """
 
+ACCOUNTS_SESSION_ACTIVITY_ENABLE = True
```
switowski: Should we use the past tense (so `ACCOUNTS_SESSION_ACTIVITY_ENABLED`) for this kind of settings? It makes more sense to me and I see it's used in other places as well (when I grep for `_ENABLE` in invenio packages, I usually find `_ENABLED` instead).

tests/test_admin.py
```diff
@@ -139,3 +141,44 @@ def test_admin_createuser(app, admin_view):
...
+    with app.test_client() as client:
+        # simulate login as user 2
+        login_user_via_view(client=client, email=users[1]['email'],
+                            password=users[1]['password'])
+        new_sid_s = session.sid_s
+        sessions = SessionActivity.query.all()
+        assert len(sessions) == 2
+        all_sid_s = [session.sid_s for session in sessions]
+        assert sorted([sid_s, new_sid_s]) == sorted(all_sid_s)
+        # and try to delete a session of another user: WORKS
```
switowski: What does this `WORKS` mean ? :P

hachreak: it's only to describe which result I should expect.
In this case I should expect that user action of session deletion end properly.

switowski: I see. Is it some kind of convention we are using in the documentation?

invenio_accounts/ext.py
hachreak: Do you think can erroneously import? 
I thought was always relative to `invenio_accounts/ext.py`

tests/test_admin.py
hachreak: Actually I don't think it could be considered a general convention, but reading the tests could not really help to understand what the result should be.
The page is not generating any exception or error in status code and if you the lines before, the check looks the same but the meaning is opposite.
And happen frequently in the past that because the code suffered a lot of refactoring, people lost the point of the test, adapting it wrongly.

switowski: I agree on documenting what is going on in a given section of tests (as we anyway don't use the `standard` approach: each test testing one thing :)
I was just curious if this is something that we use in general (and maybe this `WORKS` is a tag like `NOTE` or `TODO` or something), as I would instead just put a sentence like `and make sure that can delete a session of another user`.

PR #205, opened 2017-05-17
models: extra info addition on SessionActivity

hachreak:
* Adds extra information inside session activity. (addresses #52)

### DIFF
--- a/invenio_accounts/models.py
+++ b/invenio_accounts/models.py
@@ -135,6 +135,24 @@ class SessionActivity(db.Model, Timestamp):
 
     user = db.relationship(User, backref='active_sessions')
 
+    ip_addr = db.Column(db.String(80), nullable=True)
+    """IP address."""
+
+    country = db.Column(db.String(3), nullable=True)
+    """Country name."""
+
+    browser = db.Column(db.String(80), nullable=True)
+    """User browser."""
+
+    browser_version = db.Column(db.String(30), nullable=True)
+    """Browser version."""
+
+    operative_system = db.Column(db.String(80), nullable=True)
+    """User operative system name."""
+
+    device = db.Column(db.String(80), nullable=True)
+    """User device."""
+
     @classmethod
     def query_by_expired(cls):
         """Query to select all expired sessions."""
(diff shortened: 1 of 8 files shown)


setup.cfg
```diff
@@ -22,9 +22,6 @@
 # waive the privileges and immunities granted to it by virtue of its status
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
-[aliases]
```
jirikuncar: Why do you need to change the order?

hachreak: thanks to remember me.. I forgot to remove this change XD

setup.py
```diff
@@ -97,7 +97,9 @@
     'SQLAlchemy-Utils[ipaddress]>=0.31.0',
     'cryptography>=1.3',
     'invenio-i18n>=1.0.0b2',
+    'maxminddb-geolite2==2017.404',
```
jirikuncar: Is it necessary to pin the exact version? What if one wants to upgrade it?

invenio_accounts/models.py
```diff
@@ -135,6 +135,21 @@ class SessionActivity(db.Model, Timestamp):
 
     user = db.relationship(User, backref='active_sessions')
 
+    country = db.Column(db.String(80), nullable=True)
```
lnielsen: Perhaps we should store just an ISO code instead so we are language neutral

hachreak:
ping @lnielsen

invenio_accounts/sessions.py
```diff
@@ -42,6 +45,27 @@
 _sessionstore = LocalProxy(lambda: flask.current_app.kvsession_store)
 
 
+def _get_user_country():
+    """Get user country."""
+    user_real_ip = flask.request.headers.get(
```
lnielsen: You can assume that the real ip address is in ``request.remote_addr``. When the request is forwarded you need to use the WSGIProxyFix middleware.

hachreak: how can I mock the remote address from `flask.request.remote_addr`?

PR #208, opened 2017-05-23
global: addition of JWT

drjova:
* Adds support for JWT (jwt.io), which is enabled by default. All the
  protected endpoints ``require_api_auth`` will be automatically
  check for JWT in the headers. A token can be generated by using the
  ``{{ jwt() }}`` context processor in jinja templates and can
  be send with each API request.

### DIFF
--- a/invenio_accounts/templates/invenio_accounts/jwt.html
+++ b/invenio_accounts/templates/invenio_accounts/jwt.html
@@ -1,0 +1,24 @@
+{#
+# This file is part of Invenio.
+# Copyright (C) 2017 CERN.
+#
+# Invenio is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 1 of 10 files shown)


invenio_accounts/templates/invenio_accounts/jwt.html
```diff
@@ -0,0 +1,20 @@
...
+#
+# Invenio is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+#}
```
lnielsen: Perhaps a ``-#}`` To remove whitespace from the output?

invenio_accounts/utils.py
```diff
@@ -0,0 +1,134 @@
...
+    try:
+        return decode(
+            token,
+            current_app.config['ACCOUNTS_JWT_SECRET_KEY'],
+            algorithms=[
+                current_app.config['ACCOUNTS_JWT_ALOGORITHM']
+            ]
+        )
+    except DecodeError:
+        raise JWTDecodeError
```
lnielsen: To not mask the real error I think it would be good to either:

```
# pip install future
from future.utils import raise_from

try:
    # ...
except DecodeError as exc:
    raise_from(JWTDecodeError(), exc)
# Equivalent to Python 3 version of "raise JWTDecodeError() from exc"
```

drjova:
https://github.com/inveniosoftware/invenio-oauth2server/pull/115

PR #213, opened 2017-06-19
installation: addition of missing passlib dependency

krzysztof:
Signed-off-by: Krzysztof Nowak <k.nowak@cern.ch>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -99,6 +99,7 @@ install_requires = [
     'cryptography>=1.3',
     'invenio-i18n>=1.0.0b2',
     'maxminddb-geolite2>=2017.404',
+    'passlib>=1.7.1',
     'pyjwt>=1.5.0',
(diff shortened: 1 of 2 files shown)


krzysztof:
it's used by not specified in setup.py

lnielsen:
Is it causing troubles? I think it's specified in flask-security

krzysztof:
@lnielsen
'pwd' is new in 1.7: https://github.com/inveniosoftware/invenio-accounts/blob/1ae89f87e1bf84bd8a9d06fece740783e5bc28c7/invenio_accounts/admin.py#L38
https://pythonhosted.org/passlib/lib/passlib.pwd.html

So two issues:
- No released flask-security-fork versions specify passlib >=1.7.0
- We use this explicitly in code, shouldn't this be a dependency?

Issue #212, opened 2017-06-16
global: update to flask-security v3


lnielsen:

jacquerie:
AFAICT, this was closed by #229... or am I missing something?

lnielsen:
We're just checking during the current Invenio sprint if there's any unintended impact in other modules and then we'll make a release, but should be fine.

PR #229, opened 2017-07-27
global: encrypt_password -> hash_password

jacquerie:
Same motivation as https://github.com/inveniosoftware/flask-security-fork/pull/28, but different implementation: here I'm not deprecating `encrypt_password`, I'm outright removing it.

Consequently, I switch from `Flask-Security-Fork` to `Flask-Security>=3.0.0`, which is guaranteed to contain the commit I introduced above. The other two changes in `setup.py` are required to match the minimal version of the packages required by `Flask-Security>=3.0.0`.

### DIFF
--- a/invenio_accounts/cli.py
+++ b/invenio_accounts/cli.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2015, 2016 CERN.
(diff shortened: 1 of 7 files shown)


lnielsen:
Addresses #212

drjova:
https://github.com/inveniosoftware/invenio-oauthclient/pull/141

PR #230, opened 2017-07-28
ext: make session permanent by default

jmartinm:
Signed-off-by: Javier Martin Montull <javier.martin.montull@cern.ch>


jmartinm:
After adding this, the cookie now has expiration date.

screen shot 2017-07-28 at 11 53 45 am

PR #233, opened 2017-08-01
ext: disable remember me token generation and login from header

### LINKED ISSUE #226: RFC Remove or keep "Remember me" feature?
To remember me or not to remember me that's the question.

Currently the "Remember me" together with the session activity tracking is broken:

E.g. you have two sessions with remember me open, then from session 1 you revoke the session 2. What will happen when the browser window with session 2 is reload is that the user is logged in via the remember me token (also when a user is logged in via remember me, the session activity tracking is not created). This should not happen since you revoked the session (when you logout, the remember me token is removed)

Now the question is if we really need the remember me feature anymore? Currently we are using permanent sessions with lifetime of 31 days. I think the remember me feature is useful if you have short session lifetime (like minutes or hours), but probably not when you have long session life times.

**Proposal 1: Remove remember me**
This will solve all problems with remember me loopholes. 

**Proposal 2: Keep remember me**
We will need to monkey patch Flask-Security and Flask-Login to make it work with the session activity tracking.

Any strong opinions on either side?
@jacquerie @jmartinm This affects you related to your recent PR in oauthclient: inveniosoftware/invenio-oauthclient#134

### LINKED ISSUE #231: global: flask-login review
check if login by header/token which is enabled by default by flask-login have any security impact on Invenio.

hachreak:
* Disables remember me token generation. (closes #226)

* Disables login from headers. 

* Disables login from request. (closes #231)

### DIFF
--- a/invenio_accounts/ext.py
+++ b/invenio_accounts/ext.py
@@ -100,11 +100,12 @@ class InvenioAccounts(object):
             pass
         LoginManager._set_cookie = patch_do_nothing
 
-        # Disable loading user from header because we want to be sure we
-        # can load user only through the login form.
+        # Disable loading user from headers and object because we want to be
+        # sure we can load user only through the login form.
         def patch_reload_anonym(self, *args, **kwargs):
             self.reload_user()
         LoginManager._load_from_header = patch_reload_anonym
+        LoginManager._load_from_request = patch_reload_anonym
 
     def load_obj_or_import_string(self, value):
         """Import string or return object.
--- a/invenio_accounts/ext.py
+++ b/invenio_accounts/ext.py
@@ -32,7 +32,7 @@ import pkg_resources
 import six
 from flask import current_app, session
 from flask_kvsession import KVSessionExtension
-from flask_login import user_logged_in, user_logged_out
+from flask_login import LoginManager, user_logged_in, user_logged_out
 from flask_security import Security, changeable, recoverable, registerable, \
     utils
 from invenio_db import db
@@ -92,6 +92,13 @@ class InvenioAccounts(object):
             changeable.hash_password = hash_password
             recoverable.hash_password = hash_password
             registerable.hash_password = hash_password
+
+        # Disable remember me cookie generation as it does not work with
+        # session activity tracking (a remember me token will bypass revoking
+        # of  a session).
+        def patch_set_cookie(*args, **kwargs):
+            pass
+        LoginManager._set_cookie = patch_set_cookie
 
     def load_obj_or_import_string(self, value):
         """Import string or return object.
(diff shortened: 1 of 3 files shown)


invenio_accounts/ext.py
```diff
@@ -93,6 +93,11 @@ def monkey_patch_flask_security():
             recoverable.encrypt_password = encrypt_password
             registerable.encrypt_password = encrypt_password
 
+        # disable remember me cookie generation
```
lnielsen: Since the tests are failing, perhaps expand the comment a little (so that we can remember why it was disabled when we all are not around anymore :-):

```
Disable remember me cookie generation as it does not work with
session activity tracking (a remember me token will bypass revoking 
of  a session).
```

hachreak:
ping @lnielsen

invenio_accounts/ext.py
```diff
@@ -93,6 +93,19 @@ def monkey_patch_flask_security():
...
+        # session activity tracking (a remember me token will bypass revoking
+        # of  a session).
+        def patch_do_nothing(*args, **kwargs):
+            pass
+        LoginManager._set_cookie = patch_do_nothing
+
+        # Disable loading user from header because we want to be sure we
+        # can load user only through the login form.
+        def patch_reload_anonym(self, *args, **kwargs):
+            self.reload_user()
```
lnielsen: Just curious: Why is the ``reload_user()`` needed?

invenio_accounts/ext.py
```diff
@@ -93,6 +93,19 @@ def monkey_patch_flask_security():
             recoverable.hash_password = hash_password
             registerable.hash_password = hash_password
 
+        # Disable remember me cookie generation as it does not work with
+        # session activity tracking (a remember me token will bypass revoking
+        # of  a session).
+        def patch_do_nothing(*args, **kwargs):
+            pass
+        LoginManager._set_cookie = patch_do_nothing
```
lnielsen: Perhaps it's also worthwhile to patch ``LoginManager._clear_cookie`` since we're never setting the cookie, and thus never need to unset as well.

invenio_accounts/ext.py
```diff
@@ -93,6 +93,19 @@ def monkey_patch_flask_security():
...
+
+        # Disable loading user from header because we want to be sure we
+        # can load user only through the login form.
+        def patch_reload_anonym(self, *args, **kwargs):
+            self.reload_user()
+        LoginManager._load_from_header = patch_reload_anonym
```
lnielsen: Same thing here. I think we should likely also patch ``LoginManager._load_from_request``?

invenio_accounts/ext.py
hachreak: The cookie there is the `remember me` cookie.
I patched to be sure that it will never create that cookie.

invenio_accounts/ext.py
lnielsen: @hachreak Can you have look at if we should also patch ``_load_from_request``? I think this is the last thing missing to close #231

PR #235, opened 2017-08-03
global: disable ``remember me`` option

drjova:
* Disables ``remember me`` option from the login form.

### DIFF
--- a/invenio_accounts/ext.py
+++ b/invenio_accounts/ext.py
@@ -40,7 +40,7 @@ from passlib.registry import register_crypt_handler
 from werkzeug.utils import cached_property, import_string
 
 from invenio_accounts.forms import confirm_register_form_factory, \
-    register_form_factory
+    login_form_factory, register_form_factory
 
 from . import config
 from .datastore import SessionAwareSQLAlchemyUserDatastore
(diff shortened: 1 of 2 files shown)


invenio_accounts/ext.py
```diff
@@ -198,6 +198,9 @@ def init_app(self, app, sessionstore=None, register_blueprint=True):
                 app.extensions['security'].confirm_register_form, app
             )
 
+        app.extensions['security'].login_form = login_form_factory(
```
hachreak: Maybe it could be better to pass the [login_form as parameter](https://github.com/mattupstate/flask-security/blob/develop/flask_security/core.py#L474) when you [initialize the security app](https://github.com/inveniosoftware/invenio-accounts/blob/master/invenio_accounts/ext.py#L189)?

drjova: True, but I just followed the same pattern https://github.com/inveniosoftware/invenio-accounts/blob/master/invenio_accounts/ext.py#L193-L199

lnielsen:
This one is not needed after #233 was merged right?

hachreak:
ping @lnielsen

PR #236, opened 2017-08-03
ext: ui class

dinosk:
* ADD ui extension class to make session permanent only
      in requests from the browser.
  (blocks https://github.com/inveniosoftware/invenio-oauth2server/pull/132)
    
    Signed-off-by: Dinos Kousidis <konstantinos.kousidis@cern.ch>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -144,7 +144,7 @@ setup(
             'invenio_accounts_rest = invenio_accounts:InvenioAccountsREST',
         ],
         'invenio_base.apps': [
-            'invenio_accounts = invenio_accounts:InvenioAccounts',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -145,6 +145,7 @@
         ],
         'invenio_base.apps': [
             'invenio_accounts = invenio_accounts:InvenioAccounts',
```
lnielsen: You should remove this line, since you're calling ``super(InvenioAccountsUI, self).init_app(..)`` above

dinosk: Ah yes right

hachreak:
could you please add `InvenioAccountsUI` [here](https://github.com/inveniosoftware/invenio-accounts/blob/master/invenio_accounts/__init__.py#L58)?

PR #245, opened 2017-08-12
admin: remove one last encrypt_password

jacquerie:
Apologies for the trivial PR, but https://github.com/inveniosoftware/invenio-accounts/pull/229 missed one last use of `encrypt_password` that was introduced after I created that commit.


lnielsen:
Thanks for spotting it! :-)

PR #251, opened 2017-09-22
setup: simplekv unpin

jacquerie:
Version 0.11.2 of simplekv restored the old behavior of accepting
keys of type ``str`` in Python 2.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -104,7 +104,7 @@ install_requires = [
     'passlib>=1.7.1',
     'pyjwt>=1.5.0',
     'redis>=2.10.5',
-    'simplekv==0.10',  # FIXME: see issue simplekv#57
+    'simplekv>=0.11.2',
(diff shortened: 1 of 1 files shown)


hachreak:
https://github.com/mbr/flask-kvsession/issues/43#issuecomment-332214555

Maybe if they can release the new `flask-kvsession`, this PR is not required. :smile:

jacquerie:
> Maybe if they can release the new `flask-kvsession`, we can remove the dependency. :smile:

Yes, I prefer this solution too! So I opened a PR there: https://github.com/mbr/flask-kvsession/pull/45. Hopefully we will get a quick merge and release, and then I will amend this PR to just remove this line.

lnielsen:
Waiting for fix of build on master (which is being fixed as part of the current Invenio sprint).

PR #272, opened 2018-11-01
templates: Use SETTINGS_TEMPLATE as parent template

fenekku:
By following this pattern used by invenio-gitub,
invenio-userprofiles and so on, the parent template of
invenio_accounts/settings/security.html can be overriden
in one location.

### DIFF
--- a/invenio_accounts/ext.py
+++ b/invenio_accounts/ext.py
@@ -238,6 +238,12 @@ class InvenioAccounts(object):
             )
         )
 
+        # Follow SETTINGS_TEMPLATE pattern
+        app.config.setdefault(
+            'ACCOUNTS_SETTINGS_TEMPLATE',
+            app.config.get('SETTINGS_TEMPLATE',
+                           'invenio_accounts/settings/base.html'))
+
         config_apps = ['ACCOUNTS', 'SECURITY_']
         for k in dir(config):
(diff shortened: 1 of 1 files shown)


fenekku:
I don't think that was working for me since I had to add `ACCOUNTS_SETTINGS_TEMPLATE` to my `config.py`, but I will double check.

roksys:
@fenekku, Did you manage to check it?

invenio_accounts/ext.py
```diff
@@ -238,6 +238,12 @@ def init_config(self, app):
             )
         )
 
+        # Follow SETTINGS_TEMPLATE pattern
```
roksys: Could you set a breakpoint and check if `SETTINGS_TEMPLATE` is already initialised? Also could you do this for blueprint? We tried it locally and it seems to be set already in the blueprint.

lnielsen:
@slint and @roksys have checked it locally here by putting in some breakpoints to check it, and everything looks correctly initialized already, thus I'm closing the PR, however if you still have the issue we can try to help you debug it to see where it fails in the instance.

fenekku:
I double checked and it works :+1: . I must have fixed a mis-configuration on my part since then. Thanks for verifying this!

PR #276, opened 2018-12-04
 config: faster confirmation link expiration

topless:
- removed unnecessary devel requirement
- registered invenio-db in test_requirements
- fixed test where endpoint was registered after the first request
- closes #269


topless:
It fails travis for few python versions, I will reopen when I fix it.

PR #277, opened 2018-12-10
ext: add http headers about current user/session

### LINKED ISSUE #247: 'KVSession' object has no attribute 'sid_s'
Every now and then we get the following error:

```
'KVSession' object has no attribute 'sid_s'

Stacktrace (most recent call last):

  File "flask/app.py", line 1632, in finalize_request
    response = self.process_response(response)
  File "flask/app.py", line 1856, in process_response
    response = handler(response)
  File "invenio_accounts/sessions.py", line 116, in _commit
    delete_session(session.sid_s)
  File "werkzeug/local.py", line 347, in __getattr__
    return getattr(self._get_current_object(), name)
```

Maybe the following method (https://github.com/mbr/flask-kvsession/blob/83238b74d4e4d2ffbdfd65c1c0a00ceb4bdfd9fa/flask_kvsession/__init__.py#L89) should be called instead to avoid the attribute error?

roksys:
* Adding headers about current user/session to the http response.
  Closes #247

### DIFF
--- a/invenio_accounts/config.py
+++ b/invenio_accounts/config.py
@@ -206,3 +206,6 @@ RECAPTCHA_PUBLIC_KEY = None
 
 RECAPTCHA_PRIVATE_KEY = None
 """reCAPTCHA private key."""
+
+ACCOUNTS_USERINFO_HEADERS = False
+"""If True, add X-Session-ID and X-User-ID to the HTTP response."""
--- a/tests/test_invenio_accounts.py
+++ b/tests/test_invenio_accounts.py
@@ -13,13 +13,14 @@ from __future__ import absolute_import, print_function
 import os
 
 import pytest
+import requests
 from flask import Flask
 from flask_babelex import Babel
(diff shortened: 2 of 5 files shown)


tests/test_invenio_accounts.py
```diff
@@ -11,6 +11,7 @@
 from __future__ import absolute_import, print_function
 
 import os
+import requests
```
lnielsen: Tests are complaining about the import sorting. Can you check and repush.

invenio_accounts/config.py
```diff
@@ -206,3 +206,6 @@
 
 RECAPTCHA_PRIVATE_KEY = None
 """reCAPTCHA private key."""
+
+ACCOUNTS_INFO = True
```
lnielsen: Now, that the tests are failing can I get you to change this to ``ACCOUNTS_USERINFO_HEADERS`` and set it to False by default.

PR #278, opened 2018-12-13
config: faster confirmation link expiration

topless:
* Fixed test where endpoint was registered after the first request
* Cleanup unused declarations
* Updated assets build for example app
* Tested expiration links for password recovery and email verification
* Closes #269

### DIFF
--- a/examples/app-setup.sh
+++ b/examples/app-setup.sh
@@ -11,22 +11,10 @@ DIR=`dirname "$0"`
 cd $DIR
 export FLASK_APP=app.py
 
-# Setup app
-mkdir $DIR/instance
-
-mkdir static
-
-# Install specific dependencies
-pip install -r requirements.txt
-npm install -g node-sass@3.8.0 clean-css@3.4.19 requirejs@2.2.0 uglify-js@2.7.3
-
-# Install assets
-flask npm
-cd static
-npm install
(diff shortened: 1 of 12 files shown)


examples/app-setup.sh
```diff
@@ -11,22 +11,10 @@ DIR=`dirname "$0"`
 cd $DIR
 export FLASK_APP=app.py
 
-# Setup app
-mkdir $DIR/instance
-
-mkdir static
-
-# Install specific dependencies
-pip install -r requirements.txt
```
lnielsen: I think you need to leave this one

invenio_accounts/config.py
```diff
@@ -90,6 +90,20 @@
 SECURITY_PASSWORD_SALT = None
 """Salt for storing passwords."""
 
+SECURITY_CONFIRM_EMAIL_WITHIN = '30 minutes'
```
lnielsen: Is it possible to add a test to ensure that the link expires

examples/app-setup.sh
topless: I have a branch to test it on my travis where I have reverted these ones.

topless:
This pr is cleaned up to address only the config issue on #279

PR #288, opened 2019-11-01
ext: set kvsession data store through factory

### LINKED ISSUE #283: ext: allow more flexible KVSession storage backend via config
Currently, there are only two choices for the storage backend of Flask-KVSession, Redis or in-memory. It should be possible via a configuration variable to have a dynamically generated storage backend:

```python
# in your instance's config.py
ACCOUNTS_SESSION_STORE_FACTORY = 'my_app.session.memcached_session_store'
...


# in my_app/session.py
def memcached_session_store(app):
    from simplekv.memory.memcachestore import MemcacheStore
    return MemcacheStore(app.config['ACCOUNTS_MEMCACHED_URL']
```

mvidalgarcia:
closes #283

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -31,8 +31,10 @@ the secret key is leaked, it is imperative that a new secret key is created.
 
 Sessions
 --------
-Server-side session data is currently stored in Redis, you must therefore
-configure where your Redis instance is located using:
+Server-side session data can be saved in different data stores (e.g. Redis),
+you must therefore provide a factory that returns the KV session store object:
+
+.. autodata:: invenio_accounts.config.ACCOUNTS_SESSION_STORE_FACTORY
 
 .. autodata:: invenio_accounts.config.ACCOUNTS_SESSION_REDIS_URL
 
--- a/invenio_accounts/ext.py
+++ b/invenio_accounts/ext.py
@@ -32,7 +32,7 @@ from .datastore import SessionAwareSQLAlchemyUserDatastore
 from .hash import InvenioAesEncryptedEmail, _to_binary
 from .models import Role, User
 from .sessions import login_listener, logout_listener
-from .utils import set_session_info
+from .utils import obj_or_import_string, set_session_info
 
 
 def get_hmac(password):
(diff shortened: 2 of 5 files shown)


invenio_accounts/ext.py
```diff
@@ -248,21 +258,6 @@ def init_config(self, app):
             if any([k.startswith(prefix) for prefix in config_apps]):
                 app.config.setdefault(k, getattr(config, k))
 
-        # Set Session KV store
-        if app.config.get('ACCOUNTS_SESSION_REDIS_URL'):
```
Glignos: So since we want to keep the next version backward compatible, I would suggest to not remove the `ACCOUNTS_SESSION_REDIS_URL` and instead, if it is present then initialize it as before, but by giving priority to the `ACCOUNTS_SESSION_STORE_FACTORY`.

invenio_accounts/ext.py
```diff
@@ -53,6 +53,18 @@ def hash_password(password):
     return current_app.extensions['security'].pwd_context.hash(password)
 
 
+def load_obj_or_import_string(imp):
```
Glignos: This would look better if it got moved to the utils.py file. :)
It would also be a good idea to add both of [these](https://github.com/inveniosoftware/invenio-records-rest/blob/4932d82b922f67b3062ecb5c9ae1cf267e97d4ae/invenio_records_rest/utils.py#L60-L81) in there just in case for future use.

docs/configuration.rst
```diff
@@ -31,10 +31,10 @@ the secret key is leaked, it is imperative that a new secret key is created.
 
 Sessions
 --------
-Server-side session data is currently stored in Redis, you must therefore
-configure where your Redis instance is located using:
+Server-side session data can be saved in different data stores (e.g. Redis),
+you must therefore configure a factory that returns the store object:
```
Glignos: It would be more accurate to 'provide a factory '.
I believe that defining what kind of object we are expecting would be more helpfull 'returns the KV session store object', I'm nitpicking since we also mention it in the configuration.

Glignos:
These comments are just some improvements otherwise LGTM! :)

invenio_accounts/ext.py
```diff
@@ -205,6 +202,19 @@ def delay_security_email(msg):
         if app.config.get('ACCOUNTS_USERINFO_HEADERS'):
             request_finished.connect(set_session_info, app)
 
+        # Set Session KV store
+        if app.config.get('ACCOUNTS_SESSION_STORE_FACTORY'):
+            factory = load_obj_or_import_string(
+                app.config.get('ACCOUNTS_SESSION_STORE_FACTORY'))
+            session_kvstore = factory(app)
+        else:
+            from simplekv.memory import DictStore
```
Glignos: This block could also be provided to the `ACCOUNTS_SESSION_STORE_FACTORY ` as the default factory.

invenio_accounts/ext.py
Glignos: This function can then be replaced by `obj_or_import_string`.

invenio_accounts/ext.py
```diff
@@ -205,6 +202,19 @@ def delay_security_email(msg):
         if app.config.get('ACCOUNTS_USERINFO_HEADERS'):
             request_finished.connect(set_session_info, app)
 
+        # Set Session KV store
+        if app.config.get('ACCOUNTS_SESSION_STORE_FACTORY'):
+            factory = load_obj_or_import_string(
+                app.config.get('ACCOUNTS_SESSION_STORE_FACTORY'))
+            session_kvstore = factory(app)
```
Glignos: Before calling the factory you can also have a check before if it is indeed a function, like: `if callable(factory): ...`

drjova:
Supersede by https://github.com/inveniosoftware/invenio-accounts/pull/291

PR #290, opened 2019-11-19
setup: add Flask-Admin as an extra

### LINKED ISSUE #268: global: make Flask-Admin an extra
Installing `invenio-accounts` automatically installs `Flask-Admin`, which might not be needed for e.g. REST API-only applications. Creating an `admin` extra is `setup.py` is probably the solution to this ([example in `invenio-pidstore`](https://github.com/inveniosoftware/invenio-pidstore/blob/bc2018ece0982955f639a48f54a7ff8ddedfe5d4/setup.py#L36-L38)).

drjova:
* Closes #268

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,6 +32,9 @@ extras_require = {
     ':python_version=="2.7"': [
         'ipaddr>=2.1.11',
     ],
+    'admin': [
+        'Flask-Admin>=1.3.0',
+    ],
     'celery': [
         # Needed for building the documentation until v4.2 is released.
         'celery>=3.1.0,<4.0',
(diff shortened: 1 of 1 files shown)


drjova:
@slint what about ``BabelEx``, ``Breadcrumbs`` etc?

slint:
The crappy part about Breadcrumbs and Menu, is that they're directly imported and used as decorators in `views/security.py`, so there's no going around them unless we start doing `try: ... except ImportError: ...` and replace them with dummy decorators...

For now I'll merge, since making it possible to just not have `Flask-Admin` at all is good enough. I've created https://github.com/inveniosoftware/invenio-accounts/issues/293 for that

PR #291, opened 2019-11-20
sessions: addition of session store factory 

drjova:


drjova:
@slint if you can check the last commit, for me makes more sense to have ``None`` instead of ```''```

tests/conftest.py
```diff
@@ -51,19 +51,16 @@ def _app_factory(config=None):
         SECURITY_PASSWORD_SALT="CHANGE_ME_ALSO",
         SECURITY_CONFIRM_EMAIL_WITHIN="2 seconds",
         SECURITY_RESET_PASSWORD_WITHIN="2 seconds",
+        ACCOUNTS_SESSION_STORE_FACTORY=(
```
slint: Isn't that the default already from `config.py`?

drjova: 👍 updated

PR #292, opened 2019-11-21
sessions: stop persist anonymous

drjova:


drjova:
@slint updated

invenio_accounts/sessions.py
```diff
@@ -152,3 +154,14 @@ def default_session_store_factory(app):
             accounts_session_redis_url))
     from simplekv.memory import DictStore
     return DictStore()
+
+
+class KVSessionInterfaceWithAnonymousSessions(KVSessionInterface):
```
kprzerwa: small impact: IMO the name indicates that we store sessions for anonymous. I would suggest `KVSessionInterfaceStoreAuthenticated`

diegodelemos:
@ppanero working using latest `cookiecutter-invenio-instance`:
```console
~/invenio-accounts $ git diff
diff --git a/setup.py b/setup.py
index 0eed598..81e41f9 100644
--- a/setup.py
+++ b/setup.py
@@ -75,7 +75,7 @@ install_requires = [
      'email-validator>=1.0.5',
      'future>=0.16.0',
      'invenio-base>=1.2.2',
-    'invenio-i18n>=1.2.0',
+    'invenio-i18n<1.2.0,>=1.1.1',
      'invenio-celery>=1.1.2',
      'invenio-rest>=1.1.3',
      'maxminddb-geolite2>=2017.404',
$ cd ../ && cookiecutter cookiecutter-invenio-instance
$ cd my-site
$ cat my_site/config.py
...
# Sessions
# ========
#: Whether the instance should use SameSite cookies to store client sessions.
SESSION_COOKIE_SAMESITE = 'Strict'
...
$ ./scripts/bootstrap
$ ./scripts/setup
$ pipenv install ../invenio-accounts # the dev version
$ pip env shell
$ pip uninstall flask-kvsession -y
$ pip show flask-kvsession-invenio 
Name: Flask-KVSession-Invenio
Version: 0.6.3
...
$ ./scripts/server
```

Screenshot 2020-05-08 at 16 41 05


Note: Invenio-i18n version needs to be changed because Invenio 3.2.1 doesn't support it, but this doesn't change anything regarding Flask-KVSession integration


We need to know what we will put as the default value for SameSite cookies on sessions, in the past, we have been always choosing the most secure defaults but I am not sure for this case CC @ppanero.

PR #294, opened 2019-11-29
utils: add "register_user" function

slint:
* Adds a "register_user" function for allowing more flexible
  confirmation link sending.


slint:
Closed in favour of #295

PR #295, opened 2019-12-06
Accounts rest auth views

zzacharo:


invenio_accounts/views/__init__.py
```diff
@@ -9,5 +9,6 @@
 """Invenio-accounts views."""
 
 from .settings import blueprint
+from .rest import create_blueprint
 
-__all__ = ('blueprint', )
+__all__ = ('blueprint', 'create_blueprint')
```
slint: The naming here is a bit misguiding (since `create_blueprint` would be for the REST view). I'd say either rename (e.g. `create_rest_blueprint`) or remove and use the full path directly in `setup.py`

invenio_accounts/views/rest.py
```diff
@@ -0,0 +1,520 @@
...
+    :rtype: dict.
+    """
+    return dict(
+        id=role.id,
+        name=role.name,
+        description=role.description,
+    )
+
+
+def role_self_link(role, **kwargs):
```
slint: Not needed since we don't have a role view

setup.py
```diff
@@ -127,14 +130,17 @@
...
+            'invenio_accounts_rest_auth = invenio_accounts:InvenioAccountsREST',
         ],
         'invenio_base.apps': [
             'invenio_accounts_ui = invenio_accounts:InvenioAccountsUI',
         ],
         'invenio_base.blueprints': [
             'invenio_accounts = invenio_accounts.views.settings:blueprint',
         ],
+        'invenio_base.api_blueprints': [
+            'invenio_accounts_rest_auth = invenio_accounts.views.rest:create_blueprint',
```
slint: There's a bit of an issue here since this is a blueprint entrypoint that depends on the `rest` extras to be installed... There's a way to go around this (but it's not very pretty...):

```python
# invenio_accounts/views/__init__.py

def create_rest_blueprint(app):
    try:
        from .rest import create_blueprint
        return create_blueprint(app)
    except ImpotrError:
        warning(...)

__all__ = ('create_rest_blueprint', )


# in setup.py

...
'invenio_base.api_blueprints': [
    'invenio_accounts_rest_auth = invenio_accounts.views:create_rest_blueprint',
],
...
```

Let's discuss IRL about it.

tests/test_views_auth.py
```diff
@@ -0,0 +1,76 @@
...
+                ('email', 'user does not exist'),  # TODO: do we want this?
+            ))
+
+            # Invalid credentials
+            res = client.post(
+                url, json={'email': 'not@exists.com', 'password': '123456'})
+            assert res.status_code == 400
+            assert_error_fields(res, (
+                ('email', 'user does not exist'),
+            ))
```
slint: @zzacharo I'm writing some more tests on the rest of the views/errors. I'll keep them locally, and push after we discuss

PR #302, opened 2020-03-02
Rest api

zzacharo:


invenio_accounts/views/rest.py
```diff
@@ -0,0 +1,533 @@
...
+
+
+class SessionsListView(MethodView):
+    """View that returns the list of user sessions."""
+
+    decorators = [login_required]
+
+    def get(self, **kwargs):
+        """Return user sessions info."""
+        sessions = SessionActivity.query.filter_by(
```
lnielsen:
```python
sessions = SessionActivity.query_by_user(
        user_id=current_user.get_id()
    ).all()
```

invenio_accounts/views/rest.py
```diff
@@ -0,0 +1,533 @@
...
+            user_id=current_user.get_id())
+        results = [{
+            'created': s.created,
+            'current': s.is_current(s.sid_s),
```
lnielsen: ``SessionActivity.is_current()`` (it's a class method)

invenio_accounts/views/rest.py
```diff
@@ -0,0 +1,533 @@
...
+
+
+class SessionsItemView(MethodView):
+    """View for operations related to user session."""
+
+    decorators = [login_required]
+
+    def delete(self, sid_s=None, **kwargs):
+        """Revoke the given user session."""
+        if SessionActivity.query.filter_by(
```
lnielsen: ``SessionActivity.query_by_user(current_user.get_id()).filter_by(sid_s=sid_s).count()``

invenio_accounts/views/rest.py
```diff
@@ -0,0 +1,533 @@
...
+
+    decorators = [login_required]
+
+    confirmation_link_func = default_confirmation_link_func
+
+    post_args = {
+        'email': fields.Email(required=True, validate=[user_exists]),
+    }
+
+    def get_user(self, email=None, **kwargs):
```
lnielsen: This function is repeated at last twice, so it could be placed in a Mixin class instead to avoid duplication.

invenio_accounts/views/rest.py
```diff
@@ -0,0 +1,533 @@
...
+            login_user(user)
+
+    def success_response(self, user):
+        """Return a successful register response."""
+        return jsonify(default_user_payload(user))
+
+    @use_kwargs(post_args)
+    def post(self, **kwargs):
+        """Register a user."""
+        user = register_user(**kwargs)
```
lnielsen: The UI view has a way to extend the form so you can ask for more fields. This doesn't seem to be supported? If not, then at least an issue should be opened it.

invenio_accounts/views/rest.py
```diff
@@ -0,0 +1,533 @@
...
+                '/sessions/<sid_s>',
+                view_func=authentication_views['sessions_item'].as_view(
+                    'sessions_item'))
+    return blueprint
+
+
+class FlaskParser(FlaskParserBase):
+    """Parser to add FieldError to validation errors."""
+
+    # TODO: Add error codes to all messages (e.g. 'user-already-exists')
```
lnielsen: TODO

invenio_accounts/views/rest.py
mvidalgarcia: Do you mean that there should be more args [here](https://github.com/inveniosoftware/invenio-accounts/pull/302/files#diff-223576a60bd9ea79383ec3cf2d503ebbR274)?

invenio_accounts/views/rest.py
mvidalgarcia: I'm not sure about this. I see that `FieldError` has an optional value [`code`](https://github.com/inveniosoftware/invenio-rest/blob/master/invenio_rest/errors.py#L31) (HTTP code) but I don't know where to take it from.
OTOH, `RESTValidationError` already [sets code](https://github.com/inveniosoftware/invenio-rest/blob/master/invenio_rest/errors.py#L107) as `400`.

invenio_accounts/views/rest.py
mvidalgarcia: Issues created:
-  https://github.com/inveniosoftware/invenio-accounts/issues/304
- https://github.com/inveniosoftware/invenio-userprofiles/issues/97

invenio_accounts/views/rest.py
mvidalgarcia: Issue created:
 - https://github.com/inveniosoftware/invenio-accounts/issues/305

PR #316, opened 2020-05-08
Reset csrf token listener and add KVSession Interface.

Glignos:


ppanero:
Reopening: The Storage only for authenticated users hass issues (more in #320)

PR #318, opened 2020-05-11
Csrf update

Glignos:


invenio_accounts/sessions.py
```diff
@@ -105,6 +105,16 @@ def _commit(response=None):
         return response
 
 
+def csrf_token_reset(app, user):
+    """Connect to the user_logged_in signal to reset the csrf token.
+
+    :param app: The Flask application.
+    :param user: The :class:`invenio_accounts.models.User` instance.
+    """
+    from invenio_rest.csrf import reset_token
```
ppanero, line 114: Minor: why the import is here and not at the top? circular deps?

PR #330, opened 2020-06-01
theme: fix whitespace and sign/login consistency

lnielsen:
- Removes extra whitespace.

- Ensure consistent font size for bottom message on login/signup page.

- Ensure consistent icons for sign up button.

- Define better CSS class for sign up button.


lnielsen:
Seems to be a problem with latest released Celery :-(

PR #342, opened 2020-08-24
forms: fixes 'next' redirect in the register form and release v1.4.0a7

### LINKED ISSUE #341: views: "next" querystring parameter is not taken into account for sign-up
To replicate:

1. Run an Invenio instance
2. Go to `/admin` (or any view that requires a logged-in user)
3. You will be redirected to `/login?next=/admin`
4. If you don't have an account and want to sign-up by clicking the link you get redirected to `/signup` (notice that the `next=` parameter is not preserved)
5. After finishing the sign-up you get redirected to the homepage instead of `/admin`

Glignos:
* Closes #341

### DIFF
--- a/invenio_accounts/templates/semantic-ui/invenio_accounts/register_user.html
+++ b/invenio_accounts/templates/semantic-ui/invenio_accounts/register_user.html
@@ -48,7 +48,7 @@
 
     <div class="ui secondary segment padded">
       {{ _('Already have an account?') }}
-      <a href="{{ url_for_security('login') }}">{{ _('Log In') }}</a>
+      <a href="{{ url_for_security('login', next=request.args.get('next')) }}">{{ _('Log In') }}</a>
     </div>
   </div>
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,12 +8,14 @@
 Changes
 =======
 
-Version 1.4.0a6 (released 2020-07-28)
+Version 1.4.0a7 (released 2020-07-28)
 
 - Added Turkish translation
 - Marked strings for translation
 - Includes Semantic-UI stylings fixes.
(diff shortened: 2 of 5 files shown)


invenio_accounts/templates/semantic-ui/invenio_accounts/register_user.html
```diff
@@ -48,7 +48,7 @@ <h3 class="ui header">
 
     <div class="ui secondary segment padded">
       {{ _('Already have an account?') }}
-      <a href="{{ url_for_security('login') }}">{{ _('Log In') }}</a>
+      <a href="{{ url_for_security('login', next=request.args.get('next')) }}">{{ _('Log In') }}</a>
```
lnielsen, line 51: Minor: Shouldn't the parameter be validated before being included? Here we're directly passing through a query parameter from a client, hence an attacker should create a link and send it to a victim?

lnielsen (commented):
Please check the security implications of passing through the parameter.

invenio_accounts/templates/semantic-ui/invenio_accounts/register_user.html
slint: Indeed, probably taken from `invenio-oauthclient`, but we actually run the `get_safe_redirect_target` utility there inside the handlers... I think flask-security has some form of protection but not the same as we use

Glignos: After some further discussion with @slint we concluded that there is no improvement by additionally validating the URL through ```get_safe_redirect_target``` since the flask-security form and redirect function will be enforcing its own validator limiting the ```next``` options to subdomains. Currently, it seems that there is more effort required to change this and it doesn't seem that it's in the scope of this PR to do so.

PR #347, opened 2020-10-15
rest: add get method for logout

kprzerwa:
* fixes redirection after logout from flask-security

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -8,7 +8,7 @@
 # under the terms of the MIT License; see LICENSE file for more details.
 
 pydocstyle invenio_accounts tests docs && \
-isort -rc -c -df && \
+isort invenio_accounts tests --check-only --diff && \
 check-manifest --ignore ".travis-*" && \
 sphinx-build -qnNW docs docs/_build/html && \
 python setup.py test && \
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -6,10 +6,9 @@
 #
 # Invenio is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
-
-pydocstyle invenio_accounts tests docs && \
-isort invenio_accounts tests --check-only --diff && \
-check-manifest --ignore ".travis-*" && \
-sphinx-build -qnNW docs docs/_build/html && \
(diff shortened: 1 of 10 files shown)


ntarocco:
@slint anything against this?
The issue we have is that we need to instruct the browser to redirect to CERN logout after the Invenio logout, and it is easier to user a link `<a />` (GET request) to the endpoint then via a POST request in the React app.

slint (approved):
LGTM, and totally fine since just a feature addition :)

Also, this is basically how it's done today for the UI app

setup.py
```diff
@@ -76,7 +75,7 @@
     'future>=0.16.0',
     'invenio-base>=1.2.2',
     'invenio-i18n>=1.2.0',
-    'invenio-celery>=1.1.2',
```
ntarocco, line 79: curiosity, was this change needed?

setup.py
```diff
@@ -22,9 +22,8 @@
     'mock>=2.0.0',
     'pydocstyle>=1.0.0',
     'pytest-cov>=1.8.0',
-    'pytest-flask>=0.10.0,<1.0.0',
```
ntarocco, line 25: You can remove more deps here, double check `pytest-invenio` deps

run-tests.sh
```diff
@@ -6,9 +6,7 @@
 #
 # Invenio is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
-
-pydocstyle invenio_accounts tests docs && \
-isort -rc -c -df && \
+docker-services-cli up && \
```
ntarocco, line 9: probably you don't need this. You can have this file as in the module cookiecutter, without `docker-services-cli`:

```
python -m check_manifest --ignore ".travis-*" && \
python -m sphinx.cmd.build -qnNW docs docs/_build/html && \
python -m pytest
python -m sphinx.cmd.build -qnNW -b doctest docs docs/_build/doctest
tests_exit_code=$?
exit "$tests_exit_code"
```

setup.py
kprzerwa: 👀 mistake, thanks

ntarocco (approved):

PR #351, opened 2020-11-30
style: harmonizes login style

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-theme/issues/210

image


ppanero (approved):
Nice! One minor comment (it happens also on login and register page): Is the `Invenio)` logo centered? looks a bit left from the middle to me.

jrcastro2:
> Nice! One minor comment (it happens also on login and register page): Is the `Invenio)` logo centered? looks a bit left from the middle to me.

Nice catch! You are right it has some margin that should not apply there

PR #352, opened 2020-12-01
tests: remove explicit test dependencies

topless:
- removed dependencies introduced from pytest-invenio
- replaced backport mock with built in unittest.mock
- closes #349

PR #353, opened 2020-12-01
forms: Fixes errors rendering

### LINKED ISSUE #325: Fix errors rendering on login pages
Now: Screenshot 2020-05-15 at 14 20 49

errors should be field mapped, and removed from the general form_error message, like:
Screenshot 2020-05-15 at 14 48 31

it is broken now, because of `errormsg` hardcoded to `False`
Possible solutions:
```           
{{ render_field(form.email, icon="user icon", autofocus=True, errormsg='email' in form.errors) }}```

Since you are there, please replace all the Log in with `Sign in` phrase

### PR DESCRIPTION
**Pull request by jrcastro2**
* closes #325 

image

image

### DIFF
--- a/invenio_accounts/templates/invenio_accounts/forgot_password.html
+++ b/invenio_accounts/templates/invenio_accounts/forgot_password.html
@@ -34,7 +34,7 @@
     {# contents of this if-block can be always shown after mattupstate/flask-security#291 is released. #}
     {%- if current_user.is_anonymous %}
     <div class="panel-footer text-center">
-        <h4 class="text-muted"><a href="{{url_for('security.login')}}">{{_('Log In')}}</a>{% if security.registerable %} {{_('or')}} <a href="{{url_for('security.register')}}">{{_('Sign Up')}}</a>{% endif %}</h4>
+        <h4 class="text-muted"><a href="{{url_for('security.login')}}">{{_('Sign In')}}</a>{% if security.registerable %} {{_('or')}} <a href="{{url_for('security.register')}}">{{_('Sign Up')}}</a>{% endif %}</h4>
     </div>
     {%- endif %}
   </div>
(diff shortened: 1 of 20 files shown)

### CODE REVIEW

File: invenio_accounts/templates/invenio_accounts/forgot_password.html [hunk 1]
```diff
@@ -34,7 +34,7 @@ <h3 class="text-center panel-free-title">{{_('Reset Password')}}</h3>
     {# contents of this if-block can be always shown after mattupstate/flask-security#291 is released. #}
     {%- if current_user.is_anonymous %}
     <div class="panel-footer text-center">
-        <h4 class="text-muted"><a href="{{url_for('security.login')}}">{{_('Log In')}}</a>{% if security.registerable %} {{_('or')}} <a href="{{url_for('security.register')}}">{{_('Sign Up')}}</a>{% endif %}</h4>
+        <h4 class="text-muted"><a href="{{url_for('security.login')}}">{{_('Sign In')}}</a>{% if security.registerable %} {{_('or')}} <a href="{{url_for('security.register')}}">{{_('Sign Up')}}</a>{% endif %}</h4>
```
ppanero, line 37: Nice catch 🔥

ppanero (approved):
Really nice job! Thanks a lot for all the `Sign` fixes! 

It LGTM! Just wondering what happens if there are more than those two in `form.errors`, it seems right now they would not be shown. On the other hand, I think is safe to assume only those two can happen (and if more this form template should be updated).

**Important note** the sign up button color is not orange. **This is already fixed** in another PR, so no more work is needed here.

invenio_accounts/templates/invenio_accounts/forgot_password.html
lnielsen: Minor: We terms "Log in" and "Sign up" are correct. We should not use "Sign in".

lnielsen:
Pending merge of #352

PR #368, opened 2021-05-05
login: add possibility to disable local login

max-moser:
partially closes https://github.com/inveniosoftware/docs-invenio-rdm/issues/153

TODO:
the REST login views aren't yet affected by disabling the local login


invenio_accounts/ext.py
```diff
@@ -144,6 +147,17 @@ def init_app(self, app, sessionstore=None, register_blueprint=True):
         # Monkey-patch Flask-Security
         InvenioAccounts.monkey_patch_flask_security()
 
+        # Override Flask-Security's default login view function
+        @app.before_first_request
+        def override_login_view():
+            # this is only done before the first request, to give other
```
ntarocco, line 153: I read this comment and it makes sense.
I am just wondering why it cannot be done simply when creating this extension and not in the `before_first_request`.
Is it because otherwise we cannot be sure that the `security` extension is already loaded (issue with the extensions loading order?)

max-moser: brief testing would always give me the view function configured in `invenio_accounts.config` if not delayed until `before_first_request`
but this may have been due to another issue (the `invenio_oauthclient` not overriding the default config value); gonna investigate!

max-moser: indeed, it works after resolving that other issue, resolved!

invenio_accounts/ext.py
```diff
@@ -197,6 +210,18 @@ def delay_security_email(msg):
...
+        sec = app.extensions["security"]
+        local_login = app.config.get("ACCOUNTS_LOCAL_LOGIN_ENABLED", True)
+        x_able = sec.registerable or sec.changeable or sec.recoverable
+        if x_able and not local_login:
+            warning_message = (
+                "ACCOUNTS_LOCAL_LOGIN_ENABLED is False, while at least one "
+                "of SECURITY_REGISTERABLE, SECURITY_REGISTERABLE and "
+                "SECURITY_REGISTERABLE is True"
+            )
+            warn(warning_message, Warning, stacklevel=2)
```
max-moser, line 223: i figured, a warning would be nicer than an exception because there *might* be use cases for this?
stacklevel=2 makes the display of the warning nicer (it'll show as `init_app()`, which is nicer than `warn(...)`

invenio_accounts/views/settings.py
```diff
@@ -54,7 +54,7 @@ def init_menu():
     item = current_menu.submenu('breadcrumbs.settings')
     item.register('invenio_userprofiles.profile', _('Account'))
 
-    ### Register menu ###
+    # # # Register menu # # #
```
max-moser, line 57: this was a pycodestyle issue

setup.py
```diff
@@ -27,13 +28,13 @@
...
-        'invenio-db[versioning,mysql]>=1.0.8',
+        'invenio-db[versioning,mysql]>=1.0.9',
     ],
     'postgresql': [
-        'invenio-db[versioning,postgresql]>=1.0.8',
+        'invenio-db[versioning,postgresql]>=1.0.9',
     ],
     'sqlite': [
-        'invenio-db[versioning]>=1.0.8',
+        'invenio-db[versioning]>=1.0.9',
```
max-moser, line 37: this was required to fix the error at `apply_driver_hacks`

invenio_accounts/ext.py
```diff
@@ -197,6 +210,18 @@ def delay_security_email(msg):
...
             session_kvstore, app)
 
+        # Warn if inconsistent configuration is detected
+        sec = app.extensions["security"]
+        local_login = app.config.get("ACCOUNTS_LOCAL_LOGIN_ENABLED", True)
+        x_able = sec.registerable or sec.changeable or sec.recoverable
+        if x_able and not local_login:
+            warning_message = (
+                "ACCOUNTS_LOCAL_LOGIN_ENABLED is False, while at least one "
+                "of SECURITY_REGISTERABLE, SECURITY_REGISTERABLE and "
```
ntarocco, line 220: small mistake: `SECURITY_REGISTERABLE` is used 3 times here.

invenio_accounts/ext.py
```diff
@@ -197,6 +210,18 @@ def delay_security_email(msg):
         self.kvsession_extension = KVSessionExtension(
             session_kvstore, app)
 
+        # Warn if inconsistent configuration is detected
+        sec = app.extensions["security"]
+        local_login = app.config.get("ACCOUNTS_LOCAL_LOGIN_ENABLED", True)
+        x_able = sec.registerable or sec.changeable or sec.recoverable
```
ntarocco, line 216: Only IMO, very very very minor :)

```suggestion
        local_account_is_editable = sec.registerable or sec.changeable or sec.recoverable
```

I would also have isolated all this block in a method for readability...

invenio_accounts/views/__init__.py
```diff
@@ -2,12 +2,33 @@
...
+
+@anonymous_user_required
+@blueprint.route("/login")
+def login(*args, **kwargs):
+    """Disable login credential submission if local login is disabled."""
+    local_login_enabled = current_app.config.get(
+        "ACCOUNTS_LOCAL_LOGIN_ENABLED", True
+    )
+
+    if request.method != "GET" and not local_login_enabled:
```
ntarocco, line 27: I am not sure I understand why the `!= 'GET'` check: don't we want to return 404 also if the user goes to the web page? (maybe that's another view...)

invenio_accounts/views/settings.py
ntarocco: oh wow! :)

You can also change it as the one above `# - Register menu`

tests/test_views.py
```diff
@@ -151,6 +151,26 @@ def test_login_from_headers_disabled(app, users):
...
+        url = url_for_security('login')
+        data = {'email': email, 'password': password, 'remember': True}
+
+        res = client.post(
+            url,
+            data=data,
+            environ_base={'REMOTE_ADDR': '127.0.0.1'}
+        )
+
+        assert res.status_code != 200
```
ntarocco, line 171: Why not:
```suggestion
        assert res.status_code == 404
```
?

invenio_accounts/config.py
```diff
@@ -187,6 +190,25 @@
 SECURITY_CHANGE_URL = '/account/settings/password/'
 """URL endpoint for password change."""
 
+SECURITY_MSG_LOCAL_LOGIN_DISABLED = ("Local login is disabled.", "error")
```
ntarocco, line 193: Just wondering: why having all these messages configurable? Do you see the need for it?

ntarocco (commented):
Nice work! Looks simple to me, I have added a few comments, have a look if they make sense to you!

invenio_accounts/views/__init__.py
max-moser: no, it's the same view function that handles the login page rendering on GET and the login process on POST
if we give a 404 on GET as well, that'd override the rendering of the login page
however, we still want to show the login page when there's multiple external auth services configured, but local login disabled

PR #369, opened 2021-05-14
setup: bump versions

max-moser:


setup.py
```diff
@@ -52,6 +52,7 @@
 ]
 
 install_requires = [
+    'Flask>=1.1.4,<2.0.0',
```
max-moser, line 55: with `Flask>=1.1.0`, some tests for min reqs would fail, because of the `SECRET_KEY` apparently being set to `None`

ntarocco: Ideally we try to avoid to do this as much as possible, we moved it to `invenio-base` to avoid duplication everywhere.
I am wondering if moving `invenio-base` up, it solves the issue...

PR #376, opened 2021-09-30
tasks: add deletion of login-ips older than 30 days

### LINKED ISSUE #374: celery task: clean last login ip for ips older than 30 days

jennur:
closes #374

### DIFF
--- a/invenio_accounts/tasks.py
+++ b/invenio_accounts/tasks.py
@@ -10,12 +10,14 @@
 
 from __future__ import absolute_import, print_function
 
+from datetime import datetime
+
 from celery import shared_task
 from flask import current_app
 from flask_mail import Message
 from invenio_db import db
 
-from .models import SessionActivity
+from .models import SessionActivity, User
 from .proxies import current_accounts
 from .sessions import delete_session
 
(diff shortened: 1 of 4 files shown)


invenio_accounts/tasks.py
```diff
@@ -55,3 +57,18 @@ def clean_session_table():
...
         delete_session(sid_s=session.sid_s)
     db.session.commit()
+
+
+@shared_task
+def delete_ips():
+    """Automatically remove user.last_login_ip older than 30 days."""
+    expiration_date = datetime.utcnow() - timedelta(days=30)
+
+    users = User.query.filter(User.last_login_ip.isnot(None),
```
lnielsen, line 67: This method will load the users into the Python layer. Since we don't need the data in Python, we can execute a database UPDATE statement instead like this:  

```python
User.query.filter(
    User.last_login_ip.isnot(None), 
    User.last_login_at < expiration_date
).update({
    User.last_login_ip: None
})
User.query.filter(
    User.current_login_ip.isnot(None), 
    User.current_login_at < expiration_date
).update({
    User.current_login_ip: None
})
db.session.commit()
```

invenio_accounts/tasks.py
```diff
@@ -55,3 +57,18 @@ def clean_session_table():
     for session in sessions:
         delete_session(sid_s=session.sid_s)
     db.session.commit()
+
+
+@shared_task
+def delete_ips():
+    """Automatically remove user.last_login_ip older than 30 days."""
+    expiration_date = datetime.utcnow() - timedelta(days=30)
```
lnielsen, line 65: We need to make the time delta configurable, thus in ``config.py`` you should add something like:

```python
ACCOUNTS_RETENTION_PERIOD = timedelay(days=30)
```

Then this line you can change to:

```python
expiration_date = datetime.utcnow() - current_app.config['ACCOUNTS_RETENTION_PERIOD']
```

lnielsen (commented):
Looks good, just some minor comment...

PR #383, opened 2022-02-08
models: add version counters to users and roles

max-moser:


invenio_accounts/alembic/dfbdf43a3e96_separate_login_info.py
```diff
@@ -0,0 +1,112 @@
...
+            " current_login_at = li.current_login_at, "
+            " current_login_ip = li.current_login_ip, "
+            " last_login_at = li.last_login_at, "
+            " last_login_ip = li.last_login_ip, "
+            " login_count = li.login_count "
+            "FROM accounts_user_login_information AS li "
+            "WHERE accounts_user.id = li.user_id;"
+        )
+    elif dialect_name == "mysql":
+        # TODO untested; maybe it would be possible with sqlalchemy core?
```
max-moser, line 97: :eyes:

lnielsen: no big deal I think......downgrading rarely ever happens

lnielsen (approved):

invenio_accounts/alembic/dfbdf43a3e96_separate_login_info.py
max-moser: i dropped the untested mysql query for data retention on downgrade, as this isn't expected to be a common use case, and the information isn't precisely vital

PR #390, opened 2022-02-21
models: add user profiles as a validated json field

max-moser:
In this pull request, we introduce JSON-based fields for user profiles, which use configurable marshmallow schemas for validation of manipulation operations before their actual execution.
While this introduces additional overhead for each of the manipulation operations, we do not think that this should be a big issue because updates to user profiles are a rare operation.
Investigations showed that other Invenio modules would often use the `User` model directly rather than helpers like the datastore (e.g. to query for users), and it would thus be beneficial to not introduce extra API functions for user manipulation.

This PR enables operations like the following on users:
```python
In [6]: user.user_profile = {"full_name": "Max Moser", "username": "max-moser"}

In [7]: user.user_profile
Out[7]: {'full_name': 'Max Moser', 'username': 'max-moser'}

In [8]: user.user_profile = {"full_name": "Max Moser", "username": "max-moser", "email": "max@tuwien.ac.at"}

# ... stack trace omitted

ValueError: Validation failed: {'email': ['Unknown field.']}

In [9]: user.user_profile["email"] = "max@tuwien.ac.at"

# ... stack trace omitted

ValueError: Validation failed: {'email': ['Unknown field.']}

In [10]: user.user_profile
Out[10]: {'full_name': 'Max Moser', 'username': 'max-moser'}
```

Note that this PR does not perform any data migration from the tables defined by Invenio-UserProfiles.

### DIFF
--- a/invenio_accounts/models.py
+++ b/invenio_accounts/models.py
@@ -13,11 +13,33 @@ from datetime import datetime
 from flask import current_app, session
 from flask_security import RoleMixin, UserMixin
 from invenio_db import db
+from sqlalchemy.dialects import postgresql
 from sqlalchemy.exc import IntegrityError
+from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import validates
 from sqlalchemy_utils import IPAddressType, Timestamp
+from sqlalchemy_utils.types import JSONType
 
 from .errors import AlreadyLinkedError
+from .profiles import UserPreferenceDict, UserProfileDict
+from .utils import validate_username
+
+json_field = (
+    db.JSON()
+    .with_variant(
+        postgresql.JSONB(none_as_null=True),
+        'postgresql',
+    )
+    .with_variant(
+        JSONType(),
+        'sqlite',
+    )
+    .with_variant(
+        JSONType(),
+        'mysql',
+    )
+)
+
 
 userrole = db.Table(
     'accounts_userrole',
@@ -62,6 +84,12 @@ class User(db.Model, Timestamp, UserMixin):
 
     id = db.Column(db.Integer, primary_key=True)
 
+    _username = db.Column('username', db.String(255), unique=True)
+    """Lower-case version of the username, to assert uniqueness."""
+
+    _displayname = db.Column('displayname', db.String(255))
+    """Case-preserving version of the username."""
+
     email = db.Column(db.String(255), unique=True)
     """User email."""
 
@@ -82,6 +110,16 @@ class User(db.Model, Timestamp, UserMixin):
     version_id = db.Column(db.Integer, nullable=False)
     """Used by SQLAlchemy for optimistic concurrency control."""
 
+    _user_profile = db.Column(
+        "profile", json_field, default=lambda: dict(), nullable=True,
+    )
+    """The user profile as a JSON field."""
+
+    _preferences = db.Column(
+        "preferences", json_field, default=lambda: dict(), nullable=True,
+    )
+    """The user's preferences stored in a JSON field."""
+
     __mapper_args__ = {
         "version_id_col": version_id
     }
@@ -89,6 +127,61 @@ class User(db.Model, Timestamp, UserMixin):
     login_info = db.relationship(
         "LoginInformation", back_populates="user", uselist=False, lazy="joined"
     )
+
+    @hybrid_property
+    def username(self):
+        """Get username."""
+        return self._displayname
+
+    @username.setter
+    def username(self, username):
+        """Set username.
+
+        .. note:: The username will be converted to lowercase.
+                  The display name will contain the original version.
+        """
+        # if the username can't be validated, a ValueError will be raised
+        validate_username(username)
+        self._displayname = username
+        self._username = username.lower()
+
+    @hybrid_property
+    def user_profile(self):
+        """Get the user profile."""
+        # NOTE: accessing this property requires an initialized app for config
+        if self._user_profile is None:
+            return None
+        elif not isinstance(self._user_profile, UserProfileDict):
+            return UserProfileDict(**self._user_profile)
+
+        return self._user_profile
+
+    @user_profile.setter
+    def user_profile(self, value):
+        """Set the user profile."""
+        if value is None:
+            self._user_profile = None
+        else:
+            self._user_profile = UserProfileDict(**value)
+
+    @hybrid_property
+    def preferences(self):
+        """Get the user preferences."""
+        # NOTE: accessing this property requires an initialized app for config
+        if self._preferences is None:
+            return None
+        elif not isinstance(self._preferences, UserPreferenceDict):
+            self._preferences = UserPreferenceDict(**self._preferences)
+
+        return self._preferences
+
+    @preferences.setter
+    def preferences(self, value):
+        """Set the user preferences."""
+        if value is None:
+            self._preferences = None
+        else:
+            self._preferences = UserPreferenceDict(**value)
 
     def _get_login_info_attr(self, attr_name):
         if self.login_info is None:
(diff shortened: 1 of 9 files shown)


invenio_accounts/models.py
```diff
@@ -90,6 +121,44 @@ class User(db.Model, Timestamp, UserMixin):
...
+
+        return self._user_profile
+
+    @user_profile.setter
+    def user_profile(self, value):
+        """Set the user profile."""
+        if value is None:
+            self._user_profile = None
+        else:
+            self._user_profile = UserProfileDict(**value)
```
max-moser, line 141: when we decide to remove the old user profile table (`user.profile`), we could create a [synonym](https://docs.sqlalchemy.org/en/14/orm/mapped_attributes.html#synonyms) for the new field... and we need to move all the data from the old tables into this new JSON field (via alembic, i suppose)

invenio_accounts/models.py
```diff
@@ -90,6 +121,44 @@ class User(db.Model, Timestamp, UserMixin):
...
+        elif not isinstance(self._preferences, UserPreferenceDict):
+            self._preferences = UserPreferenceDict(**self._preferences)
+
+        return self._preferences
+
+    @user_preferences.setter
+    def user_preferences(self, value):
+        """Set the user preferences."""
+        if value is None:
+            self._preferences = None
```
max-moser, line 158: maybe we should set this to `{}` rather than `None`?

ntarocco (commented):
From a quick review, it looks a nice implementation to me! 👍🏼 

Just a thought: what about enforcing that any implementation of the profile/prefs should inherit from `ValidatedDict`?

invenio_accounts/profiles/classes.py
```diff
@@ -0,0 +1,96 @@
...
+            if self._schema is not None:
+                # without schema, we basically revert to a normal dictionary
+                # with more overhead
+                self._schema.load(data)
+        except ValidationError as error:
+            raise ValueError(f"Validation failed: {error}")
+
+    def _try_op(self, op_name, *args, **kwargs):
+        """Try the named operation and see if it violates the schema."""
+        data = {**self}
```
zzacharo, line 38: Is the `{**self}` syntax deep copying the "self" object?

invenio_accounts/profiles/classes.py
```diff
@@ -0,0 +1,96 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
max-moser: no, this was just the first thing that came to my mind because `utils.py` was already "taken" a level above...
a different name would indeed be better, what would you prefer?
`utils.py` (the `profiles` namespacing should make its purpose somewhat clear), `dicts.py`, or something else?

on a side note, i'm also not 100% happy with `profiles`, because it's also used for (the for now rudimentary) preferences, and might be used for other JSON fields...

invenio_accounts/models.py
max-moser: right now, the property setter makes sure that any value other than `None` is converted into a `UserPreferenceDict` (by calling the the latter's constructor with the deconstructed values), i.e. something like `user.user_profile = {"username": "max"}` (with a normal `dict`) is already fine
this is unnecessary overhead if the assigned value already is a `UserPreferenceDict`, but i think these assignments aren't happening so often that this would make a real difference...

if we rewrote the setter like the following (and the getter similarly):
```python
        if value is None:
            value = {}
        self._user_profile = UserProfileDict(**value)
```
we would have the benefit that we don't need `if user.user_profile is not None` checks in the code, before accessing its keys

max-moser:
i've tried making the JSON database columns (`user_profile`, `preferences`) use the `ValidatedDictionary` subclasses rather than the `dict` class, but my rudimentary experiments with customizing the column types/type decorators didn't show much success (especially with the per-dialect variants) until i thought a getter/setter-based approach would suit our use case well enough...

with the setters transforming passed dict-like data structures into proper `ValidatedDictionary` objects, and the getters also ensuring that the returned value is of this type, one would have to try pretty hard to get something other than a `ValidatedDictionary` out of the fields (even though it's still saved as a normal dict/JSON column)
if this is a pattern that we might want to use more commonly, it could make sense to investigate further how to make such fields easier to set (i.e. just with a db.Column, and without special getters and setters)

invenio_accounts/models.py
```diff
@@ -82,6 +103,16 @@ class User(db.Model, Timestamp, UserMixin):
     version_id = db.Column(db.Integer, nullable=False)
     """Used by SQLAlchemy for optimistic concurrency control."""
 
+    _user_profile = db.Column(
+        "user_profile", json_field, default=lambda: dict(), nullable=True,
```
lnielsen, line 107: Perhaps change name to from ``user_profile`` to ``profile`` since it's already on the user class or is it going to conflict with an existing property?
```suggestion
        "profile", json_field, default=lambda: dict(), nullable=True,
```

lnielsen (approved):
Looks great. Really my only comment is we should think twice if ``username`` should be in the profile or outside (like email). I think it depends on if we want users to login with a username or not.

invenio_accounts/models.py
max-moser: true. there's no reason that i see right now why the DB column couldn't be called `profile` already; it's just the property name on the model class that'd collide with the `profile` relationship to the current `invenio-userprofile` model

invenio_accounts/models.py
```diff
@@ -90,6 +121,44 @@ class User(db.Model, Timestamp, UserMixin):
...
     )
 
+    @hybrid_property
+    def user_profile(self):
+        """Get the user profile."""
+        # NOTE: accessing this property requires an initialized app for config
+        if self._user_profile is None:
+            return None
+        elif not isinstance(self._user_profile, UserProfileDict):
+            self._user_profile = UserProfileDict(**self._user_profile)
```
max-moser, line 131: it has to be seen if this setting of the column value causes issues down the road (currently thinking of the post-commit hooks in users-resources, since you're absolutely not allowed to do anything that causes SQL statements to be issued after a session's been committed)
i could imagine that this change doesn't actually cause any net changes, and should thus be fine
otherwise, we might have to take the (probably minimal) performance hits and rewrite this line to always convert the stored dict to a `UserProfileDict` rather than caching it:
```python
return UserProfileDict(**self._user_profile)
```

max-moser:
done; i've lifted the `username` up into the user model rather than the profile JSON to make it easier and more reliable for querying.
my thoughts here are that usernames are common enough to justify a column in the user table... if the system building on top of the invenio framework doesn't want to use usernames, this field can still be ignored
sure, it could be considered wasteful in this scenario, but i don't think minimalism is the main concern here for now

invenio_accounts/models.py
```diff
@@ -62,6 +83,9 @@ class User(db.Model, Timestamp, UserMixin):
 
     id = db.Column(db.Integer, primary_key=True)
 
+    username = db.Column(db.String(255), unique=True)
```
lnielsen, line 86: I think we have to do the username like in [UserProfiles](https://github.com/inveniosoftware/invenio-userprofiles/blob/master/invenio_userprofiles/models.py#L52-L75).

Basically we have two columns 1) case-preserving username used for display purposes and 2) a lower case version of a username with a unique constraint. This ensures we cannot have two usernames: ``lnielsen`` and ``LNIELSEN`` as they would both be the same username, but that we can preserve the case depending on how the user prefers to write it.

invenio_accounts/models.py
max-moser: i've reworked this line to be on the safer side, as i don't think that the performance hit is too dramatic here (the user profiles shouldn't be large enough to make copying & validating the dictionaries have a big impact, and then it only happens when the user's profile property is accessed)

PR #394, opened 2022-02-28
User preferences schema

max-moser:
Add the `visibility` and `email_visibility` settings to the default user preferences schema, as per discussion on https://github.com/inveniosoftware/invenio-users-resources/pull/15.
Also use a fallback value for `_user_id` in the session if it's not available.

### DIFF
--- a/invenio_accounts/sessions.py
+++ b/invenio_accounts/sessions.py
@@ -52,6 +52,11 @@ def add_session(session=None):
         is used. The object is expected to have a dictionary entry named
         ``"_user_id"`` and a field ``sid_s``
     """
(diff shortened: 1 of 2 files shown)


invenio_accounts/sessions.py
```diff
@@ -52,6 +52,9 @@ def add_session(session=None):
         is used. The object is expected to have a dictionary entry named
         ``"_user_id"`` and a field ``sid_s``
     """
+    if '_user_id' not in session and 'user_id' in session:
+        session['_user_id'] = session['user_id']
```
max-moser, line 56: not sure if this is just symptom treatment, but it fixes the `KeyError` that i've posted in Discord for me

PR #406, opened 2022-04-20
signals: add datastore pre_commit and post_commit

ppanero:
- closes https://github.com/inveniosoftware/invenio-communities/issues/536

### DIFF
--- a/invenio_accounts/proxies.py
+++ b/invenio_accounts/proxies.py
@@ -8,8 +8,10 @@
 
 """Helper proxy to the state object."""
 
-from flask import current_app
+from flask import current_app, g
 from werkzeug.local import LocalProxy
+
+from .api import DBUsersChangeHistory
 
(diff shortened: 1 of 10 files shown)


invenio_accounts/proxies.py
```diff
@@ -22,3 +24,15 @@
 current_datastore = LocalProxy(
     lambda: current_app.extensions['security'].datastore)
 """Proxy to the current Flask-Security user datastore."""
+
+
+def get_db_change_history():
+    """Proxy funtion to db change history."""
```
lnielsen, line 30: I would add a quick note here something along the lines:

> The "g" object is local to the Flask application context. An app context is pushed on every 
request or CLI command. This means that the users change history is automatically cleared with on tear down of the app context. This matches up with the database session is also tied to the app context.

```

lnielsen (approved):

PR #428, opened 2022-11-08
user: add confirmation to user_create

utnapischtim:
* with this parameter it is possible to create a new user and set
  the email confirmation in one step

### DIFF
--- a/invenio_accounts/cli.py
+++ b/invenio_accounts/cli.py
@@ -8,6 +8,7 @@
 
 """Command Line Interface for accounts."""
 
+from datetime import date
 from functools import wraps
 
 import click
@@ -46,9 +47,10 @@ def roles():
 @click.argument("email")
 @click.password_option()
 @click.option("-a", "--active", default=False, is_flag=True)
+@click.option("-c", "--confirm", default=False, is_flag=True)
 @with_appcontext
 @commit
-def users_create(email, password, active):
+def users_create(email, password, active, confirm):
     """Create a user."""
     kwargs = dict(email=email, password=password, active="y" if active else "")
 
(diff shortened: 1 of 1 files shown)


invenio_accounts/cli.py
```diff
@@ -57,6 +59,8 @@ def users_create(email, password, active):
     if form.validate():
         kwargs["password"] = hash_password(kwargs["password"])
         kwargs["active"] = active
+        if confirm:
+            kwargs["confirmed_at"] = date.today()
```
ppanero, line 63: To be consistent with other timestampts stored in the DB (e.g. created/updated) it should contain the h/m/s too.
```suggestion
            kwargs["confirmed_at"] = datetime.utcnow()
```

invenio_accounts/cli.py
```diff
@@ -8,6 +8,7 @@
 
 """Command Line Interface for accounts."""
 
+from datetime import date
```
ppanero, line 11:
```suggestion
from datetime import datetime
```

ppanero (commented):
Some minor comments, once fixed it can be merged IMO. I know @avivace will love this ;)

invenio_accounts/cli.py
utnapischtim: should we use `datetime.utcnow.isoformat()` as in https://github.com/inveniosoftware/invenio-saml/blob/1aca07d56ebec7dedd02b15ba7c7192840751493/invenio_saml/handlers.py#L69 ?

further: https://github.com/inveniosoftware/invenio-saml/pull/31#discussion_r1003742387

ppanero: I would personally keep it as a datetime and then let the DB handle representations (imagine someone does not want ISO format for some reason). That would also make it consistent with the ORM. In [records](https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/models.py#L31) we are using `utcnow` as default.

utnapischtim: ok, then i go with `datetime.utcnow()` 
i will reference this decision also in the invenio-saml PR. so we could do the same there too

avivace:
Cool! Let's just remember to update the quickstart docs to mention the new flag

ppanero (approved):
LGTM! Thanks for the work on this!

ntarocco (approved):

PR #429, opened 2022-12-01
register: option to avoid sending e-mail

ntarocco:
* adds a new option to fine tune when to send the registration e-mail and not relying only on the global security configuration.

### DIFF
--- a/invenio_accounts/utils.py
+++ b/invenio_accounts/utils.py
@@ -151,7 +151,7 @@ def default_confirmation_link_func(user):
     return token, _generate_token_url(endpoint, token)
 
 
-def register_user(_confirmation_link_func=None, **user_data):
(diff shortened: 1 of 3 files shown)


invenio_accounts/utils.py
```diff
@@ -151,7 +151,7 @@ def default_confirmation_link_func(user):
     return token, _generate_token_url(endpoint, token)
 
 
-def register_user(_confirmation_link_func=None, **user_data):
+def register_user(_confirmation_link_func=None, send_register_msg=True, **user_data):
```
zzacharo, line 154: I guess this is needed to be able to manually disable the email send? Related to invenio-oauthclient contrib login?

zzacharo (approved):

PR #434, opened 2023-02-06
migrate to flask babel

utnapischtim:
- migrate: flask_babelex replaced by invenio_i18n
- refactor: remove unused statements

- NOTES:
  - to make this testable it is necessary to release invenio-i18n, flask-security-invenio and invenio-theme

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,11 @@
 Changes
 =======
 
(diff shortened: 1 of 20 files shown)


tmorrell:
This is a fun one. pydocstyle doesn't want a blank line after the dosctring, but black wants a blank line because it's the start of a function definition.

PR #438, opened 2023-03-21
preferences: add support for preferences.locale

TLGINO:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2069

### DIFF
--- a/invenio_accounts/models.py
+++ b/invenio_accounts/models.py
@@ -12,6 +12,7 @@
 from datetime import datetime
 
 from flask import current_app, session
+from flask_babel import refresh
 from flask_security import RoleMixin, UserMixin
 from invenio_db import db
 from sqlalchemy.dialects import postgresql
@@ -150,6 +151,14 @@ class User(db.Model, Timestamp, UserMixin):
             "email_visibility",
             current_app.config.get("ACCOUNTS_DEFAULT_EMAIL_VISIBILITY", "restricted"),
         )
+        preferences.setdefault(
+            "locale",
+            current_app.config.get("BABEL_DEFAULT_LOCALE", "en"),
+        )
+        preferences.setdefault(
+            "timezone",
+            current_app.config.get("BABEL_DEFAULT_TIMEZONE", "Europe/Zurich"),
+        )
         super().__init__(*args, **kwargs)
         self.user_profile = user_profile
         self.preferences = preferences
@@ -212,6 +221,7 @@ class User(db.Model, Timestamp, UserMixin):
             self._preferences = None
         else:
             self._preferences = UserPreferenceDict(**value)
+            refresh()
 
     def _get_login_info_attr(self, attr_name):
         if self.login_info is None:
--- a/invenio_accounts/profiles/schemas.py
+++ b/invenio_accounts/profiles/schemas.py
@@ -8,7 +8,8 @@
 
 """Schemas for user profiles and preferences."""
 
-
+import pytz
+from flask import current_app
 from invenio_i18n import lazy_gettext as _
 from marshmallow import Schema, ValidationError, fields
 
@@ -19,6 +20,23 @@ def validate_visibility(value):
         raise ValidationError(
             message=str(_("Value must be either 'public' or 'restricted'."))
         )
+
+
+def validate_locale(value):
+    """Check if the value is a valid locale."""
+    locales = current_app.extensions["invenio-i18n"].get_locales()
+    locales = [locale.language for locale in locales]
+
+    if value not in locales:
+        raise ValidationError(message=str(_("Value must be a valid locale.")))
+    current_app.config["BABEL_DEFAULT_LOCALE"] = value
+
+
+def validate_timezone(value):
+    """Check if the value is a valid timezone."""
+    if value not in pytz.all_timezones:
+        raise ValidationError(message=str(_("Value must be a valid timezone.")))
+    current_app.config["BABEL_DEFAULT_TIMEZONE"] = value
 
 
 class UserProfileSchema(Schema):
@@ -33,3 +51,5 @@ class UserPreferencesSchema(Schema):
 
     visibility = fields.String(validate=validate_visibility)
     email_visibility = fields.String(validate=validate_visibility)
+    locale = fields.String(validate=validate_locale)
+    timezone = fields.String(validate=validate_timezone)
--- a/tests/test_schemas.py
+++ b/tests/test_schemas.py
@@ -1,0 +1,44 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2023 CERN.
+#
+# Invenio is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Test invenio-accounts schemas."""
+
+import pytest
+from marshmallow import ValidationError
+
+from invenio_accounts.profiles.schemas import (
+    validate_locale,
+    validate_timezone,
+    validate_visibility,
+)
+
+
+def test_validate_visibility():
+    """Test validate_visibility."""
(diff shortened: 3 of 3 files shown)


invenio_accounts/config.py
```diff
@@ -312,6 +312,9 @@
 ACCOUNTS_USER_PREFERENCES_SCHEMA = UserPreferencesSchema()
 """The schema to use for validation of the user preferences."""
 
+ACCOUNTS_DEFAULT_LOCALE_PREFERENCES = "ch"
```
ntarocco, line 315:
```suggestion
ACCOUNTS_DEFAULT_LOCALE_PREFERENCES = "en_GB"
```

invenio_accounts/models.py
```diff
@@ -150,6 +150,10 @@ def __init__(self, *args, **kwargs):
             "email_visibility",
             current_app.config.get("ACCOUNTS_DEFAULT_EMAIL_VISIBILITY", "restricted"),
         )
+        preferences.setdefault(
+            "preferences_locale",
+            current_app.config.get("ACCOUNTS_DEFAULT_LOCALE_PREFERENCES", "ch"),
```
ntarocco, line 155:
```suggestion
            current_app.config.get("ACCOUNTS_DEFAULT_LOCALE_PREFERENCES", "en_GB"),
```

invenio_accounts/profiles/schemas.py
```diff
@@ -33,3 +41,4 @@ class UserPreferencesSchema(Schema):
 
     visibility = fields.String(validate=validate_visibility)
     email_visibility = fields.String(validate=validate_visibility)
+    preferences_locale = fields.String(validate=validate_preferences_locale)
```
ntarocco, line 44: Shall we remove the `preferences` prefix everywhere?
```suggestion
    locale = fields.String(validate=validate_locale)
```

invenio_accounts/models.py
```diff
@@ -150,6 +150,10 @@ def __init__(self, *args, **kwargs):
             "email_visibility",
             current_app.config.get("ACCOUNTS_DEFAULT_EMAIL_VISIBILITY", "restricted"),
         )
+        preferences.setdefault(
+            "preferences_locale",
```
ntarocco, line 154:
```suggestion
            "locale",
```

invenio_accounts/profiles/schemas.py
```diff
@@ -21,6 +21,14 @@ def validate_visibility(value):
         )
 
 
+def validate_preferences_locale(value):
+    """Check if the value is a valid locale."""
+    if False:  # TODO upddate this
```
ntarocco, line 26: we need to find a reliable way to validate a locale string, from a quick check it is not straightforward :)

invenio_accounts/config.py
```diff
@@ -312,6 +312,12 @@
 ACCOUNTS_USER_PREFERENCES_SCHEMA = UserPreferencesSchema()
 """The schema to use for validation of the user preferences."""
 
+ACCOUNTS_DEFAULT_LOCALE_PREFERENCES = "en_GB"
```
ntarocco, line 315: Shall we use `BABEL_DEFAULT_LOCALE` and `BABEL_DEFAULT_TIMEZONE` instead?

invenio_accounts/profiles/schemas.py
```diff
@@ -21,6 +23,25 @@ def validate_visibility(value):
...
 
+def validate_locale(value):
+    """Check if the value is a valid locale."""
+    locales = current_app.extensions["invenio-i18n"].get_locales()
+    locales = [locale.language for locale in locales]
+    locales.append("en_GB")
+    if value not in locales:
+        raise ValidationError(message=str(_("Value must be a valid locale.")))
+    current_app.config["BABEL_DEFAULT_LOCALE"] = value
+    refresh()
```
ntarocco, line 34: I would not add a `refresh` inside the validate method (same below).
You might want to do that when the preference is changed instead. [This](https://github.com/inveniosoftware/invenio-accounts/blob/master/invenio_accounts/models.py#L209) might be a good candidate, only when it actually really changed. There could be other better places, to be checked :)

invenio_accounts/profiles/schemas.py
```diff
@@ -21,6 +23,25 @@ def validate_visibility(value):
         )
 
 
+def validate_locale(value):
+    """Check if the value is a valid locale."""
+    locales = current_app.extensions["invenio-i18n"].get_locales()
+    locales = [locale.language for locale in locales]
```
ntarocco, line 29: I tried this in one env, and I get:
```
In [2]: locales = current_app.extensions["invenio-i18n"].get_locales()
In [3]: locales
Out[3]: [Locale('en')]
In [4]: locales = [locale.language for locale in locales]
In [5]: locales
Out[5]: ['en']
```
I don't think `locale.language` is correct. I would use a `set()` and not a `list`.
I would not add the default `locales.append("en_GB")`: you might have an instance that has only one configured locale, e.g. Italian :)

invenio_accounts/profiles/schemas.py
```diff
@@ -21,6 +23,25 @@ def validate_visibility(value):
...
+    if value not in locales:
+        raise ValidationError(message=str(_("Value must be a valid locale.")))
+    current_app.config["BABEL_DEFAULT_LOCALE"] = value
+    refresh()
+
+
+def validate_timezone(value):
+    """Check if the value is a valid timezone."""
+    if value not in pytz.all_timezones:
+        raise ValidationError(message=str(_("Value must be a valid timezone.")))
```
ntarocco, line 40: do you need to `cast` to `str`? same above
```suggestion
        raise ValidationError(message=_("Value must be a valid timezone."))
```

ntarocco (commented):
Can you add a small unit test for the validate methods?

invenio_accounts/profiles/schemas.py
ntarocco: I would:
1. understand if `flask-babel` is doing something particular with the variable `BABEL_DEFAULT_LOCALE`, e.g. setting some global context in flask
2. eventually, change [the func](https://github.com/inveniosoftware/invenio-i18n/blob/master/invenio_i18n/ext.py#L170) in `invenio-i18n` (pseudocode):

```diff
- langs = [self.babel.default_locale]
- for lang, dummy_title in current_app.config.get("I18N_LANGUAGES", []):
    - langs.append(Locale.parse(lang))
+ # is the python obj Locale implpementing the magic methods to compare, to avoid to add duplicates in a `set`?
+ locales = set()
+ # the following should be a locale, not a language only (`en_GB`, not `en`)
+ default_locale = current_app.config.get("BABEL_DEFAULT_LOCALE", self.babel.default_locale)
+ for ... loop
```

The question is: if I set language `en`, what will be the locale? `en_GB` or `en_US`, for example?

tests/test_schemas.py
```diff
@@ -0,0 +1,43 @@
...
+        validate_locale("en")
+        with pytest.raises(ValidationError):
+            validate_locale("invalid")
+
+
+def test_validate_timezone(app):
+    """Test validate_timezone."""
+    with app.app_context():
+        validate_timezone("Europe/Vienna")
+        validate_timezone("Europe/London")
```
alejandromumo, line 41: suggestion: validate our default language + locale (`en / Europe(Zurich)` from what I understood from the code)

Pipfile
```diff
@@ -0,0 +1,11 @@
+[[source]]
```
kpsherva, line 1: leftovers? we don't use pipenv inside modules

alejandromumo (dismissed):

alejandromumo:
Tests are failing, not sure if related with this PR or changes in `invenio-db`

kpsherva (approved):

PR #439, opened 2023-04-11
datastore: add update_role func 

TLGINO:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2186

### DIFF
--- a/invenio_accounts/api.py
+++ b/invenio_accounts/api.py
@@ -1,32 +1,57 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2022 CERN.
+# Copyright (C) 2022-2023 CERN.
 #
 # Invenio is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
 
 """API objects for Invenio Accounts."""
 
-from collections import defaultdict
+
+class Session:
+    """Session object for DB Users change history."""
+
+    def __init__(self):
+        """Constructor."""
+        self.updated_users = set()
+        self.updated_roles = set()
+        self.deleted_users = set()
+        self.deleted_roles = set()
 
 
 class DBUsersChangeHistory:
     """DB Users change history storage."""
 
     def __init__(self):
-        """constructor."""
-        # the keys are going to be the sessions, the values are going to be
-        # the sets of dirty/deleted models
-        self.updated_users = defaultdict(lambda: list())
-        self.updated_roles = defaultdict(lambda: list())
-        self.deleted_users = defaultdict(lambda: list())
-        self.deleted_roles = defaultdict(lambda: list())
+        """Constructor."""
+        self.sessions = {}
 
-    def _clear_dirty_sets(self, session):
-        """Clear the dirty sets for the given session."""
+    def _get_session(self, session_id):
+        """Returns or creates a session for a concrete session id."""
+        return self.sessions.setdefault(session_id, Session())
+
+    def add_updated_user(self, session_id, user_id):
+        """Adds a user to the updated users list."""
+        session = self._get_session(session_id)
+        session.updated_users.add(user_id)
+
+    def add_updated_role(self, session_id, role_id):
+        """Adds a role to the updated roles list."""
+        session = self._get_session(session_id)
+        session.updated_roles.add(role_id)
+
+    def add_deleted_user(self, session_id, user_id):
+        """Adds a user to the deleted users list."""
+        session = self._get_session(session_id)
+        session.deleted_users.add(user_id)
+
+    def add_deleted_role(self, session_id, role_id):
+        """Adds a role to the deleted roles list."""
+        session = self._get_session(session_id)
+        session.deleted_roles.add(role_id)
+
+    def clear_dirty_sets(self, session):
+        """Removes session object."""
         sid = id(session)
-        self.updated_users.pop(sid, None)
-        self.updated_roles.pop(sid, None)
-        self.deleted_users.pop(sid, None)
-        self.deleted_roles.pop(sid, None)
+        self.sessions.pop(sid, None)
--- a/invenio_accounts/datastore.py
+++ b/invenio_accounts/datastore.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2015-2022 CERN.
+# Copyright (C) 2015-2023 CERN.
 #
 # Invenio is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
@@ -10,6 +10,7 @@
 
 from flask_security import SQLAlchemyUserDatastore
 
+from .models import Role
 from .proxies import current_db_change_history
 from .sessions import delete_user_sessions
 from .signals import datastore_post_commit, datastore_pre_commit
@@ -34,10 +35,27 @@ class SessionAwareSQLAlchemyUserDatastore(SQLAlchemyUserDatastore):
         datastore_pre_commit.send(session=self.db.session)
         super().commit()
         datastore_post_commit.send(session=self.db.session)
+        current_db_change_history.clear_dirty_sets(self.db.session)
 
     def mark_changed(self, sid, uid=None, rid=None):
         """Save a user to the changed history."""
         if uid:
-            current_db_change_history.updated_users[sid].append(uid)
+            current_db_change_history.add_updated_user(sid, uid)
         elif rid:
-            current_db_change_history.updated_roles[sid].append(uid)
+            current_db_change_history.add_updated_role(sid, rid)
+
+    def update_role(self, role):
+        """Updates roles."""
+        role = self.db.session.merge(role)
+        self.mark_changed(id(self.db.session), rid=role.id)
+        return role
+
+    def create_role(self, **kwargs):
+        """Creates and returns a new role from the given parameters."""
+        role = super().create_role(**kwargs)
+        self.mark_changed(id(self.db.session), rid=role.id)
+        return role
+
+    def find_role_by_id(self, role_id):
+        """Fetches roles searching by id."""
+        return self.role_model.query.filter_by(id=role_id).one_or_none()
--- a/invenio_accounts/models.py
+++ b/invenio_accounts/models.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2015-2022 CERN.
+# Copyright (C) 2015-2023 CERN.
 # Copyright (C) 2022 KTH Royal Institute of Technology
 #
 # Invenio is free software; you can redistribute it and/or modify it
@@ -9,6 +9,7 @@
 
 """Database models for accounts."""
 
+import uuid
 from datetime import datetime
 
 from flask import current_app, session
@@ -52,7 +53,7 @@ userrole = db.Table(
     ),
     db.Column(
         "role_id",
-        db.Integer(),
+        db.String(80),
         db.ForeignKey("accounts_role.id", name="fk_accounts_userrole_role_id"),
     ),
 )
@@ -64,13 +65,16 @@ class Role(db.Model, Timestamp, RoleMixin):
 
     __tablename__ = "accounts_role"
 
-    id = db.Column(db.Integer(), primary_key=True)
+    id = db.Column(db.String(80), primary_key=True, default=str(uuid.uuid4()))
 
     name = db.Column(db.String(80), unique=True)
     """Role name."""
 
     description = db.Column(db.String(255))
     """Role description."""
+
+    is_managed = db.Column(db.Boolean(), default=True, nullable=False)
+    """True when the role is managed by Invenio, and not externally provided."""
 
     # Enables SQLAlchemy version counter
     version_id = db.Column(db.Integer, nullable=False)
--- a/tests/test_invenio_accounts.py
+++ b/tests/test_invenio_accounts.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2015-2018 CERN.
+# Copyright (C) 2015-2023 CERN.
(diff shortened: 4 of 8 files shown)


invenio_accounts/alembic/8f11b75e0995_change_accountsrole_primary_key_to_.py
ppanero, line 1: The file is fully commented and contains a TODO. Should it be reviewed?

invenio_accounts/api.py
```diff
@@ -8,25 +8,60 @@
 
 """API objects for Invenio Accounts."""
 
-from collections import defaultdict
+
+class Session:
+    """Session object for DB Users change history."""
+
+    def __init__(self):
+        self.updated_users = []
```
ppanero, line 16: (unless I am missing something) all the operations performed on this attributes are additions and check for existance (`is in...`). In that case, a set() would perform better. WDYT?

invenio_accounts/api.py
```diff
@@ -8,25 +8,60 @@
...
+        self.sessions = {}
+
+    def _get_session(self, session_id):
+        if session_id not in self.sessions:
+            self.sessions[session_id] = Session()
+        return self.sessions[session_id]
+
+    def add_updated_user(self, session_id, user_id):
+        session = self._get_session(session_id)
+        if user_id not in session.updated_users:
```
ppanero, line 38: if we change the objects from list to sets we do not need to perform a check, since the set contains unique values. So we would just call `add`.

```python
>>> test = set()
>>> test.add(1)
>>> test
{1}
>>> test.add(1)
>>> test
{1}
```

The same applies in other functions.

invenio_accounts/api.py
```diff
@@ -8,25 +8,60 @@
...
+
+class Session:
+    """Session object for DB Users change history."""
+
+    def __init__(self):
+        self.updated_users = []
+        self.updated_roles = []
+        self.deleted_users = []
+        self.deleted_roles = []
+        self.indexed = False
```
ppanero, line 20: where are `indexed` and `invalidated_cache` accessed? cant find on this pr might be in another linked one?

invenio_accounts/models.py
```diff
@@ -64,14 +65,17 @@ class Role(db.Model, Timestamp, RoleMixin):
 
     __tablename__ = "accounts_role"
 
-    id = db.Column(db.Integer(), primary_key=True)
+    id = db.Column(db.String(80), primary_key=True, default=str(uuid.uuid4()))
```
ppanero, line 68: Can use `UUIDType`, see for example here https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/models.py#L62

ppanero (commented):
Nice work going to the very roots of db hehe! 🚀  few questions and suggestions.

invenio_accounts/alembic/8f11b75e0995_change_accountsrole_primary_key_to_.py
TLGINO: I'd discussed this with @ntarocco and the consensus was that this task would be tackled in another issue at a later date

invenio_accounts/models.py
```diff
@@ -52,7 +53,7 @@
     ),
     db.Column(
         "role_id",
-        db.Integer(),
+        db.String(80),
```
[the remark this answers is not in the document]
kpsherva: same question

invenio_accounts/profiles/schemas.py
```diff
@@ -24,10 +24,12 @@ def validate_visibility(value):
 
 def validate_locale(value):
     """Check if the value is a valid locale."""
-    locales = current_app.extensions["invenio-i18n"].get_locales()
-    locales = [locale.language for locale in locales]
+    locales = current_app.extensions.get("invenio-i18n")
```
ntarocco, line 27: Why this change? `i18n` should be one base module always available.

invenio_accounts/api.py
ntarocco, line 20: are these 2 properties used elsewhere?

invenio_accounts/api.py
```diff
@@ -8,25 +8,60 @@
...
-        self.deleted_users = defaultdict(lambda: list())
-        self.deleted_roles = defaultdict(lambda: list())
-
-    def _clear_dirty_sets(self, session):
-        """Clear the dirty sets for the given session."""
+        """Constructor."""
+        self.sessions = {}
+
+    def _get_session(self, session_id):
+        if session_id not in self.sessions:
```
ntarocco, line 32: you can probably replace all these `if` with:
```
self.sessions.setdefault(session_id, Session())
```

invenio_accounts/api.py
```diff
@@ -8,25 +8,60 @@
...
+        ):
+            del self.sessions[session_id]
+
+    def clear_dirty_sets(self, session):
         sid = id(session)
-        self.updated_users.pop(sid, None)
-        self.updated_roles.pop(sid, None)
-        self.deleted_users.pop(sid, None)
-        self.deleted_roles.pop(sid, None)
+        if sid in self.sessions:
```
ntarocco, line 66: you can replace this with:
```
self.sessions.pop(sid, None)
```

invenio_accounts/profiles/schemas.py
TLGINO: either invenio-communities or invenio-access has tests which make a call to this method and crash as there is apparently no instance of `invenio-i18n`

invenio_accounts/api.py
ppanero, line 20: In the demo it was shown that the attributes are updated from invenio-access and invenio-communities. I would suggest having methods inside the `Session` class e.g.:

```py
def index():
     self.indexed = True
```

To encapsulate the handling of the flags. If we update from other modules we might end up in a situation where we do not know exactly who is updating the flags. E.g. avoid deadlocks.

invenio_accounts/api.py
```diff
@@ -8,25 +8,60 @@
 
 """API objects for Invenio Accounts."""
 
-from collections import defaultdict
+
+class Session:
```
ppanero, line 12: From what @zzacharo mentioned in the call:

I've checked uwsgi docs, shared mem is not thread safe (sometimes they do magic with the GIL). But in this case is not, so we might need to make the whole session object thread safe (e.g. locks)?

invenio_accounts/alembic/8f11b75e0995_change_accountsrole_primary_key_to_.py
ppanero: That is okay for me, however while it means that we can have new instances, this will block the upgrade of existing ones no?

cc @jrcastro2

jrcastro2: Right, we will try to fix this in the same PR, otherwise we should uncomment it.

invenio_accounts/api.py
jrcastro2: I see ... This is meant to be used by signal custom handlers and they look like they are thread safe (see [here docs](https://docs.python.org/3.4/library/signal.html#execution-of-python-signal-handlers)).

invenio_accounts/api.py
jrcastro2: Yes, in [`invenio-users-resources`](https://github.com/inveniosoftware/invenio-users-resources/pull/75/files#diff-2dbb78424a207aae3daba71b31a89e424293633e65090f2239f68dcfe5a756d9R66) custom signal handler to take care of the indexing and in [`invenio-communities`](https://github.com/inveniosoftware/invenio-communities/pull/945/files#diff-6c07ffce9d2dd97c912da89fd4869787bdb49626a3fa278d466117cc8e489941R130) to invalidate the cache.

jrcastro2: Good idea! thanks :+1:

invenio_accounts/models.py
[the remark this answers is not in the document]
jrcastro2: Ping @TLGINO , he might know

invenio_accounts/profiles/schemas.py
ntarocco: Then, IMO, we should add `i18n` as dependency on the tests

invenio_accounts/alembic/f2522cdd5fcd_change_accountsrole_primary_key_to_.py
```diff
@@ -0,0 +1,115 @@
...
+revision = "f2522cdd5fcd"
+down_revision = "eb9743315a9d"
+branch_labels = ()
+depends_on = None
+
+
+def upgrade():
+    """Upgrade database."""
+    # Drop primary key and all foreign keys
+    op.execute("ALTER TABLE accounts_role DROP CONSTRAINT pk_accounts_role CASCADE")
```
ntarocco, line 25: Let's discuss Alembic stuff IRL. It is preferable to avoid using direct queries, and use `op` instead.

invenio_accounts/api.py
ntarocco: > I see ... This is meant to be used by signal custom handlers and they look like they are thread safe (see [here docs](https://docs.python.org/3.4/library/signal.html#execution-of-python-signal-handlers)).

I am not sure that I understand this very well. There is nothing preventing the usage of this `Session` in other ways.
Why do we need to store different `session_id`? Let's chat IRL.

invenio_accounts/api.py
```diff
@@ -8,25 +8,68 @@
...
+    def __init__(self):
+        """Constructor."""
+        self.updated_users = set()
+        self.updated_roles = set()
+        self.deleted_users = set()
+        self.deleted_roles = set()
+        self.indexed = False
+        self.invalidated_cache = False
+
+    def index(self):
```
ntarocco, line 24: as discussed, remove these 2 methods, and add a `finalizer` to invenio-accounts/datastore -> commit method, after the signals.

invenio_accounts/api.py
ntarocco: @ppanero we double checked, [blinker](https://blinker.readthedocs.io/en/stable/#blinker-documentation) is thread-safe.

invenio_accounts/models.py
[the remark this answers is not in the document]
ntarocco: This was copied from the `name` field, which is 80 chars. It worked until now, that's why we went with the same.

ntarocco (commented):
For me, can be merged after agreed changes and alembic fix

invenio_accounts/alembic/f2522cdd5fcd_change_accountsrole_primary_key_to_.py
ppanero, line 1: Filename

```diff
- invenio_accounts/alembic/f2522cdd5fcd_change_accountsrole_primary_key_to_.py
+ invenio_accounts/alembic/f2522cdd5fcd_change_accountsrole_primary_key_to_string.py 
```

invenio_accounts/api.py
ppanero: Discussed with @jrcastro2 IVL, since signals are sync and the cleanup is done by `clear_dirty_sets` can this two flags be removed? are they used?

invenio_accounts/datastore.py
```diff
@@ -34,10 +35,27 @@ def commit(self):
...
         """Save a user to the changed history."""
         if uid:
-            current_db_change_history.updated_users[sid].append(uid)
+            current_db_change_history.add_updated_user(sid, uid)
         elif rid:
-            current_db_change_history.updated_roles[sid].append(uid)
+            current_db_change_history.add_updated_role(sid, rid)
+
+    def update_role(self, role):
+        """Merge roles."""
```
ppanero, line 48: docstring update

invenio_accounts/datastore.py
```diff
@@ -34,10 +35,27 @@ def commit(self):
...
+        return role
+
+    def create_role(self, **kwargs):
+        """Creates and returns a new role from the given parameters."""
+        role = super().create_role(**kwargs)
+        self.mark_changed(id(self.db.session), rid=role.id)
+        return role
+
+    def find_role_by_id(self, role_id):
+        """Merge roles."""
```
ppanero, line 60: docstring update

invenio_accounts/models.py
```diff
@@ -64,14 +65,17 @@ class Role(db.Model, Timestamp, RoleMixin):
...
-    id = db.Column(db.Integer(), primary_key=True)
+    id = db.Column(db.String(80), primary_key=True, default=str(uuid.uuid4()))
 
     name = db.Column(db.String(80), unique=True)
     """Role name."""
 
     description = db.Column(db.String(255))
     """Role description."""
 
+    is_managed = db.Column(db.Boolean(name="is_managed"), default=True, nullable=False)
```
ppanero, line 76: Does boolean need a name?

```suggestion
    is_managed = db.Column(db.Boolean(), default=True, nullable=False)
```

tests/test_invenio_accounts.py
```diff
@@ -136,6 +136,27 @@ def test_datastore_rolecreate(app):
         assert 1 == Role.query.filter_by(name="superuser").count()
 
 
+def test_datastore_update_role(app):
+    """Test update role."""
+    ds = app.extensions["invenio-accounts"].datastore
+
+    with app.app_context():
```
ppanero, line 143: Is the context manager needed, or does the `app` fixture already push an app ctx?
```suggestion

```

tests/test_invenio_accounts.py
```diff
@@ -136,6 +136,27 @@ def test_datastore_rolecreate(app):
...
+        assert 1 == Role.query.filter_by(name="superuser").count()
+
+        r1 = ds.update_role(
+            Role(id="1", name="megauser", description="updated description")
+        )
+        ds.commit()
+        r2 = ds.find_role("megauser")
+        assert r1 == r2
+        assert r2.description == "updated description"
+        assert 1 == Role.query.filter_by(name="megauser").count()
```
ppanero, line 157: I would check that superuser does not exist anymore, to verify is an update not an addition.

```python
assert 0 == Role.query.filter_by(name="superuser").count()
```

ppanero (commented):
Overall check the copyright headers, they need 2023 updates :) Nice work! 🚀

PR #447, opened 2023-07-20
models: added moderation fields

alejandromumo:
closes https://github.com/inveniosoftware/invenio-users-resources/issues/92


alejandromumo:
MAJOR: the alembic recipe's **real** down revision is the mergepoint "f9843093f686" (invenio-access). However tests fail in that case. The current down revision is "f2522cdd5fcd" (invenio-accounts) and tests are passing

zzacharo (approved):

PR #452, opened 2023-08-15
models: added default verified status.

alejandromumo:
closes https://github.com/inveniosoftware/invenio-requests/issues/344

### DIFF
--- a/invenio_accounts/config.py
+++ b/invenio_accounts/config.py
@@ -334,3 +334,6 @@ ACCOUNTS_USERNAME_RULES_TEXT = _(
 
 .. note:: Used for both form help text and for form validation error.
 """
+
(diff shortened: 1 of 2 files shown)


invenio_accounts/config.py
```diff
@@ -334,3 +334,6 @@
 
 .. note:: Used for both form help text and for form validation error.
 """
+
+ACCOUNTS_DEFAULT_VERIFIED_STATUS = False
```
alejandromumo, line 338: We discussed having a default for the verified status of users.

Some instances might prefer to have users verified by default.

zzacharo:
```suggestion
ACCOUNTS_DEFAULT_USERS_VERIFIED = True
```

I would enable it by default as it might be relevant only for zenodo

zzacharo (approved):

PR #463, opened 2023-11-07
ext: add user to db history before each flush

yashlamba:
(Potentially) Fixes https://github.com/zenodo/rdm-project/issues/487


yashlamba:
Finalised approach: https://github.com/inveniosoftware/invenio-accounts/pull/464

PR #472, opened 2024-01-22
sessions: check for request before accessing session

kpsherva:


invenio_accounts/sessions.py
```diff
@@ -132,7 +132,7 @@ def delete_session(sid_s):
     # Remove entries from sessionstore
     _sessionstore.delete(sid_s)
     # Find and remove the corresponding SessionActivity entry
-    if "_impersonator_id" not in session:
+    if request and "_impersonator_id" not in session:
```
zzacharo, line 135: Should we add this check also here https://github.com/inveniosoftware/flask-security-fork/blob/b99556579f4bbc9d1425e79d1f9f0d115e5baf6a/flask_security/utils.py#L59 ?

kpsherva: in the case you've pointed out we already have to be in the request context to impersonate the user (discussed during standup)

PR #475, opened 2024-02-12
Models: Domain statistics and model change history tracking fixes

lnielsen:
-  Fixes an issue where a newly uncommitted user would not yet have an id assigned, and hence would cause None value to be added to a list of user ids that needed to be reindexed, causing the indexing to fail due to transaction rollback.
- Add change tracking of domains to reindex on updates.
- Add task to compute domain statistics on a regular basis.
- closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2575

### Checklist

Ticks in all boxes and 🟢 on all GitHub actions status checks are required to merge:

- [x] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [x] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [x] I've added relevant test cases.
- [ ] I've added relevant documentation.
- [x] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/best-practices/i18n/) (for relevant code).
- [x] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines (for relevant code).
- [x] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines (for relevant code).
- [x] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines (for relevant code).
- [x] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [x] I've NOT included third-party code (copy/pasted source code or new dependencies).

**Third-party code**

If you've added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect](https://github.com/orgs/inveniosoftware/teams/architects/members).

**Reminder**

By using GitHub, you have already agreed to the [GitHub’s Terms of Service](https://help.github.com/articles/github-terms-of-service/#6-contributions-under-repository-license) including that:

1. You license your contribution under the same terms as the current repository’s license.
2. You agree that you have the right to license your contribution under the current repository’s license.

### DIFF
--- a/invenio_accounts/datastore.py
+++ b/invenio_accounts/datastore.py
@@ -12,6 +12,7 @@ from datetime import datetime
 
 from flask import current_app
 from flask_security import SQLAlchemyUserDatastore, user_confirmed
+from sqlalchemy.orm import joinedload
 
 from .models import Domain, Role, User
 from .proxies import current_db_change_history
@@ -27,10 +28,9 @@ class SessionAwareSQLAlchemyUserDatastore(SQLAlchemyUserDatastore):
         now = datetime.utcnow()
         user.blocked_at = None
         user.verified_at = now
+        user.active = True
         if user.confirmed_at is None:
             user.confirmed_at = now
-        if not user.active:
-            user.active = True
         return True
 
     def block_user(self, user):
@@ -38,8 +38,7 @@ class SessionAwareSQLAlchemyUserDatastore(SQLAlchemyUserDatastore):
         now = datetime.utcnow()
         user.blocked_at = now
         user.verified_at = None
-        if user.active:
-            user.active = False
+        user.active = False
         delete_user_sessions(user)
         return True
 
@@ -113,7 +112,11 @@ class SessionAwareSQLAlchemyUserDatastore(SQLAlchemyUserDatastore):
 
     def find_domain(self, domain):
         """Find a domain."""
-        return Domain.query.filter_by(domain=domain).one_or_none()
+        return (
+            Domain.query.filter_by(domain=domain)
+            .options(joinedload(Domain.category_name))
+            .one_or_none()
+        )
 
     def create_domain(self, domain, **kwargs):
         """Create a new domain."""
--- a/invenio_accounts/datastore.py
+++ b/invenio_accounts/datastore.py
@@ -10,7 +10,8 @@
 
 from datetime import datetime
 
-from flask_security import SQLAlchemyUserDatastore
+from flask import current_app
+from flask_security import SQLAlchemyUserDatastore, user_confirmed
 
 from .models import Domain, Role, User
 from .proxies import current_db_change_history
@@ -45,11 +46,11 @@ class SessionAwareSQLAlchemyUserDatastore(SQLAlchemyUserDatastore):
     def activate_user(self, user):
         """Activate a unconfirmed/deactivated/blocked user."""
         res = super().activate_user(user)
-        if res:
-            user.blocked_at = None
-            if user.confirmed_at is None:
-                user.confirmed_at = datetime.utcnow()
-        return True
+        user.blocked_at = None
+        if user.confirmed_at is None:
+            user.confirmed_at = datetime.utcnow()
+            user_confirmed.send(current_app._get_current_object(), user=user)
+        return res
 
     def deactivate_user(self, user):
         """Deactivate a  user.
--- a/invenio_accounts/datastore.py
+++ b/invenio_accounts/datastore.py
@@ -8,9 +8,11 @@
 
 """Session-aware datastore."""
 
+from datetime import datetime
+
 from flask_security import SQLAlchemyUserDatastore
 
-from .models import Domain, Role
+from .models import Domain, Role, User
 from .proxies import current_db_change_history
 from .sessions import delete_user_sessions
 from .signals import datastore_post_commit, datastore_pre_commit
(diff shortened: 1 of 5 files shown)


invenio_accounts/datastore.py
```diff
@@ -19,14 +23,46 @@
 class SessionAwareSQLAlchemyUserDatastore(SQLAlchemyUserDatastore):
     """Datastore which deletes active session when a user is deactivated."""
 
+    def verify_user(self, user):
```
lnielsen, line 26: Moved here from Invenio-users-resources to treat these action consistently.

invenio_accounts/datastore.py
```diff
@@ -19,14 +23,46 @@
...
+        delete_user_sessions(user)
+        return True
+
+    def activate_user(self, user):
+        """Activate a unconfirmed/deactivated/blocked user."""
+        res = super().activate_user(user)
+        user.blocked_at = None
+        if user.confirmed_at is None:
+            user.confirmed_at = datetime.utcnow()
+            user_confirmed.send(current_app._get_current_object(), user=user)
```
lnielsen, line 53: Needs to send signal to run the code which takes care of registering the domain.

invenio_accounts/datastore.py
```diff
@@ -37,29 +73,39 @@ def commit(self):
         datastore_post_commit.send(session=self.db.session)
         current_db_change_history.clear_dirty_sets(self.db.session)
 
-    def put(self, model):
```
lnielsen, line 40: Put was being used for both roles and users, but the ``mark_changed()`` call only handled users, and in addition the model object only have an id if it was committed which is not always the case.

invenio_accounts/datastore.py
```diff
@@ -19,14 +23,46 @@
...
 
+    def verify_user(self, user):
+        """Verify a user."""
+        now = datetime.utcnow()
+        user.blocked_at = None
+        user.verified_at = now
+        if user.confirmed_at is None:
+            user.confirmed_at = now
+        if not user.active:
+            user.active = True
```
Samk13, line 34:
```suggestion
        user.confirmed_at = user.confirmed_at or now
        user.active = True # Directly set to True, assuming verification implies activation
```

invenio_accounts/tasks.py
```diff
@@ -73,3 +74,77 @@ def delete_ips():
...
+                "num_blocked": blocked,
+                "updated": now,
+            }
+        )
+        i += 1
+        # Commit batches of 500 updates
+        if i % 500 == 0:
+            db.session.commit()
+    if i % 500 != 0:
+        db.session.commit()
```
Samk13, line 150:
```suggestion
    domain_updates = stmt.all()

    # Commit batches of 500 updates
    batch_size = 500
    now = datetime.utcnow()

    # Process updates in batches
    for i in range(0, len(domain_updates), batch_size):
        with db.session.begin_nested():  # Use nested transactions for safety
            for domain, users, active, inactive, confirmed, verified, blocked in domain_updates[i:i + batch_size]:
                db.session.query(Domain).filter(Domain.domain == domain).update({
                    "num_users": users,
                    "num_active": active,
                    "num_inactive": inactive,
                    "num_confirmed": confirmed,
                    "num_verified": verified,
                    "num_blocked": blocked,
                    "updated": now,
                })
        db.session.commit() # Commit after each batch
```
no need for a final commit check outside the loop as it's handled within the loop.

Samk13: Should we try, expect any errors while unpacking the domain_updates, such as changes in the structure that could lead to a ValueError, or others?

invenio_accounts/datastore.py
Samk13: Your concern is valid If there's a chance that `user.confirmed_at ` could be mistakenly set to a falsy value after it's been correctly set to a timestamp once, then the `or` expression would indeed be dangerous because it would then reset `user.confirmed_at ` to `now`.
 But with normal use and validation applies to this db column, assuming `user.confirmed_at `  is either None (not set) or a valid timestamp (meaning it has a truthy value), the `or` expression to me is safe and does exactly what's intended unless there are other scenarios I am not aware of!

invenio_accounts/models.py
```diff
@@ -564,12 +561,15 @@ class Domain(db.Model, Timestamp):
...
     """Organisation associated with domain."""
 
-    org = db.relationship("DomainOrg", back_populates="domains")
+    org = db.relationship("DomainOrg", backref="domains")
 
     # spammer, mail-provider, organisation, company
     category = db.Column(db.Integer(), db.ForeignKey(DomainCategory.id), nullable=True)
     """Category of domain."""
 
+    category_name = db.relationship("DomainCategory", backref="domains")
```
kpsherva, line 570: alembic recipe is missing, was not committed ?

invenio_accounts/datastore.py
lnielsen: Updated and also the block _user is updated with the same issue.

invenio_accounts/models.py
```diff
@@ -564,12 +561,15 @@ class Domain(db.Model, Timestamp):
     org_id = db.Column(db.Integer(), db.ForeignKey(DomainOrg.id), nullable=True)
     """Organisation associated with domain."""
 
-    org = db.relationship("DomainOrg", back_populates="domains")
+    org = db.relationship("DomainOrg", backref="domains")
```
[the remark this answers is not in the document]
lnielsen: Since the category can be null I would just leave it blank, and let instances define values themselves.......my guess is most instances won't need it....an admin interface to edit would be needed at some point :-)

invenio_accounts/tasks.py
lnielsen: Updated. I would simply let the full task fail if there's an error and let the error logging take care of the rest. Main thing I can think of would be transaction conflicts or similar.

PR #482, opened 2024-06-27
cli: add command for domain create

utnapischtim:


invenio_accounts/cli.py
```diff
@@ -154,3 +163,18 @@ def users_deactivate(user):
...
         click.secho('User "%s" was already deactivated.' % user, fg="yellow")
+
+
+@domains.command("create")
+@click.argument("domain")
+@with_appcontext
+def domains_create(domain):
+    """Create domain."""
+    try:
+        domain_category = DomainCategory.create(domain.lower())
```
utnapischtim, line 174: question: is the lower really necessary? should it be like that?

slint: The column is actually named `label` so I would imagine it's the human-friendly name. IMHO it shouldn't have  to be lowercase, but it also doesn't matter since the FK is an integer

slint (approved):

PR #487, opened 2024-08-08
revert: commit f9a8a85

utnapischtim:
* it introduced circular dependency with invenio-users-resources. the
  feature will be moved to another package.


run-tests.sh
```diff
@@ -44,7 +44,7 @@ fi
 python -m check_manifest
 python -m setup extract_messages --output-file /dev/null
 python -m sphinx.cmd.build -qnN docs docs/_build/html
-eval "$(docker-services-cli up --db ${DB:-postgresql} --search ${SEARCH:-opensearch} --cache ${CACHE:-redis} --env)"
```
zzacharo, line 47: this should stay no? Otherwise, we cannot test search no?

utnapischtim: it was an addition of the commit i will revert. i thought to remove all what was added in the https://github.com/utnapischtim/invenio-accounts/commit/f9a8a8519b99abf0543886e230dfd73f870ad611

zzacharo: I think we can keep that... Basically instead of `git revert <commit>` you can just push a new one with only removing the necessary parts

utnapischtim: if we add that to the tests then a container will be started which is not used. that consumes resources which are unnecessary 
since it was not there before in my opinion we should save resources

zzacharo (approved):

kpsherva:
which version do we need to yank?

zzacharo:
> which version do we need to yank?

5.1.0

PR #492, opened 2024-09-18
security: filter out master_session if it's set to `None`

### LINKED ISSUE #491: `None` values in the list of sessions breaks the `security.html` template
Sometimes it happened to me that the `security.html` page (with the user's sessions) breaks.
Especially, this happens when impersonating users in the system through the administration panel.
Testing has shown that some of the `session` objects being iterated over in the `security.html` templates are `None`.
This breaks at the datetime formatting filters running over the session objects: https://github.com/inveniosoftware/invenio-accounts/blob/master/invenio_accounts/templates/invenio_accounts/settings/security.html#L37

max-moser:
Closes #491 

The resulting security page when impersonating users:
image

It may look a little strange, but i consider that better than a 500 internal server error :smile:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,11 @@
 Changes
 =======
 
+Version 5.1.2 (released 2024-09-19)
+
+- setup: bump minimum flask-security-invenio dependency
+- security: handle missing value for current session
+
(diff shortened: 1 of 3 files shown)


ntarocco (approved):
LGTM and thank you!

PR #500, opened 2024-11-06
bug: add handler for handling changed domain

carlinmack:
:heart: Thank you for your contribution!

(Part of) Close https://github.com/inveniosoftware/invenio-app-rdm/issues/2777

### Description

> Please describe briefly your pull request.

[We call `mark_changed`](https://github.com/carlinmack/invenio-accounts/blob/commit-domain/invenio_accounts/domains.py#L23) but we did not check that it handled Domains

Calling flush as similarly done in the [pre-commit hook](https://github.com/inveniosoftware/invenio-users-resources/blob/master/invenio_users_resources/records/hooks.py#L31)

Also remove setting of role.id as the added flush() will add the ID from DB

### Checklist

Ticks in all boxes and 🟢 on all GitHub actions status checks are required to merge:

- [ ] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [ ] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [ ] I've added relevant test cases.
- [ ] I've added relevant documentation.
- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).
- [ ] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [ ] I've NOT included third-party code (copy/pasted source code or new dependencies).
    * If you have added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect](https://github.com/orgs/inveniosoftware/teams/architects/members).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines.

**Reminder**

By using GitHub, you have already agreed to the [GitHub’s Terms of Service](https://help.github.com/articles/github-terms-of-service/#6-contributions-under-repository-license) including that:

1. You license your contribution under the same terms as the current repository’s license.
2. You agree that you have the right to license your contribution under the current repository’s license.

### DIFF
--- a/invenio_accounts/datastore.py
+++ b/invenio_accounts/datastore.py
@@ -14,6 +14,7 @@ from datetime import datetime
 from flask import current_app
 from flask_security import SQLAlchemyUserDatastore, user_confirmed
 from invenio_db import db
+from sqlalchemy import inspect
 from sqlalchemy.orm import joinedload
 
 from .models import Domain, Role, User
@@ -75,10 +76,11 @@ class SessionAwareSQLAlchemyUserDatastore(SQLAlchemyUserDatastore):
 
     def mark_changed(self, sid, uid=None, rid=None, model=None):
         """Save a user to the changed history."""
-        # needed so that we have the id from the DB on the model
-        self.db.session.flush()
-
         if model:
+            # add the ID to the model from the DB if needed
+            if not inspect(model).persistent:
+                self.db.session.flush()
+
             if isinstance(model, User):
                 current_db_change_history.add_updated_user(sid, model.id)
             elif isinstance(model, Role):
--- a/invenio_accounts/datastore.py
+++ b/invenio_accounts/datastore.py
@@ -75,11 +75,16 @@ class SessionAwareSQLAlchemyUserDatastore(SQLAlchemyUserDatastore):
 
     def mark_changed(self, sid, uid=None, rid=None, model=None):
         """Save a user to the changed history."""
+        # needed so that we have the id from the DB on the model
+        self.db.session.flush()
+
         if model:
             if isinstance(model, User):
                 current_db_change_history.add_updated_user(sid, model.id)
             elif isinstance(model, Role):
                 current_db_change_history.add_updated_role(sid, model.id)
+            elif isinstance(model, Domain):
+                current_db_change_history.add_updated_domain(sid, model.id)
         elif uid:
             # Deprecated - use model param instead (still used in e.g.
             # UserFixture pytest-invenio)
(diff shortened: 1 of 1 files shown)


invenio_accounts/datastore.py
```diff
@@ -118,6 +115,14 @@ def find_domain(self, domain):
...
         )
 
+    def find_domain_by_id(self, domain_id):
+        """Find a domain by id."""
+        return (
+            Domain.query.filter_by(id=domain_id)
+            .options(joinedload(Domain.category_name))
+            .one_or_none()
+        )
+
```
slint, line 125:
```suggestion
    def find_domain(self, domain_or_id):
        """Find a domain by it's value or ID."""
        if isinstance(domain_or_id, str):
            if domain_or_id.isdigit():
                clause = Domain.id == int(domain_or_id)
            else:
                clause = Domain.domain == domain_or_id
        elif isinstance(domain_or_id, int):
            clause = Domain.id == domain_or_id
        else:
            raise ValueError("...")
        return (
            Domain.query.filter(clause)
            .options(joinedload(Domain.category_name))
            .one_or_none()
        )
```

minor: IMHO this class should try to be as flexible/helpful as possible. Thus, enhancing the existing `find_domain` method to do a bit of inspection to accommodate more use cases is fine and keeps the complexity from spreading to other places.

invenio_accounts/datastore.py
```diff
@@ -118,6 +115,14 @@ def find_domain(self, domain):
             .one_or_none()
         )
 
+    def find_domain_by_id(self, domain_id):
+        """Find a domain by id."""
+        return (
+            Domain.query.filter_by(id=domain_id)
```
utnapischtim, line 121: sqlalchemy: please use the 2.0 style: db.session.query(Domain).filter_by(...)

invenio_accounts/datastore.py
```diff
@@ -73,11 +73,16 @@ def commit(self):
 
     def mark_changed(self, sid, uid=None, rid=None, model=None):
         """Save a user to the changed history."""
+        # needed so that we have the id from the DB on the model
+        self.db.session.flush()
```
lnielsen, line 77: @slint Can you have a look at the impact of having a flush here.

slint: I see we very often tiptoe around flushes, and get them in implicitly just by chance because we perform another `SELECT` query down the line, e.g.:

```python
domain = Domain.create(...)

# ...in another function...

user = UserAggregate.get(...)  # <-- this triggered a flush, so our `domain`
                               #     from above now has a `domain.id`
```

So I think it's better to be explicit about it, which is what calling `mark_changed` also implies (i.e. that something will be persisted to the DB).

@carlinmack is also checking this change on tests across other repos to see if there are wrong expectations.

invenio_accounts/datastore.py
```diff
@@ -75,11 +76,17 @@ def commit(self):
 
     def mark_changed(self, sid, uid=None, rid=None, model=None):
         """Save a user to the changed history."""
+        # add the ID to the model from the DB if needed
+        if not inspect(model).persistent:
+            self.db.session.flush()
+
```
carlinmack, line 82: perhaps I move above into the `if model:` block

zubeydecivelek (approved):
LGTU! 🚀  Peer reviewed with @jrcastro2

PR #501, opened 2024-11-08
model: add method to check if key is present in user_profile

jrcastro2:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/193

### DIFF
--- a/invenio_accounts/models.py
+++ b/invenio_accounts/models.py
@@ -20,7 +20,7 @@ from invenio_db import db
 from sqlalchemy import func
 from sqlalchemy.dialects import postgresql
 from sqlalchemy.exc import IntegrityError
-from sqlalchemy.ext.hybrid import Comparator, hybrid_property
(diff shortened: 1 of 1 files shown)


invenio_accounts/models.py
```diff
@@ -271,6 +271,16 @@ def preferences(self, value):
...
 
+    @hybrid_method
+    def has_key_in_profile(self, key):
+        """Check if the given key exists in the user_profile."""
+        return self._user_profile.get(key) is not None
+
+    @has_key_in_profile.expression
+    def has_key_in_profile(cls, key):
+        """SQL expression for filtering by presence of a key in the user_profile."""
+        return func.jsonb_extract_path_text(cls._user_profile, key).isnot(None)
```
ntarocco, line 282: Postgres only?

jrcastro2:
> Is this PR really needed? No other ways?

Just thought we would prefere to filter it out when searching from the DB, not crucial, so we can simply do it in python. Closing this one as it's not anymore required. Thanks for the comment!

PR #505, opened 2024-11-29
Fix domain

jrcastro2:


carlinmack (approved):

slint (approved):
LGTM, thanks for the quickfix 🙏

invenio_accounts/__init__.py
```diff
@@ -55,7 +55,7 @@
 from .ext import InvenioAccounts, InvenioAccountsREST, InvenioAccountsUI
 from .proxies import current_accounts
 
-__version__ = "5.1.6"
+__version__ = "5.1.5"
```
slint, line 58: Oof, this should be `5.1.7`

PR #520, opened 2025-06-26
fix: pkg_resources DeprecationWarning

utnapischtim:
* UserWarning: pkg_resources is deprecated as an API. See
  https://setuptools.pypa.io/en/latest/pkg_resources.html. The
  pkg_resources package is slated for removal as early as 2025-11-30.
  Refrain from using this package or pin to Setuptools<81.


tmorrell (approved):
Good modernization.

PR #523, opened 2025-07-03
fix: find_spec raise ValueError

utnapischtim:
* find_spec raises ValueError if the spec is not found


tmorrell (approved):
This seems like the best match for the existing behavior

PR #528, opened 2025-07-29
cli: don't allow role changes on externally managed

### LINKED ISSUE #445: cli: don't allow to change membership of externally managed group
- if the group is externally managed, we should not be able to manage it internally (via cli)
- if the membership can be managed internally, we might run into inconsistencies of access

edivalentinitu:
- as per issue https://github.com/inveniosoftware/invenio-accounts/issues/445, added new condition in role related CLI operations to restrict the possibility of membership change on externally managed groups.


fenekku:
@utnapischtim to take a look for v14

fenekku (approved):
Thanks!

PR #533, opened 2025-09-30
fix(chore): DeprecationWarning stdlib

utnapischtim:
* datetime.datetime.utcnow() is deprecated and scheduled for removal in
  a future version. Use timezone-aware objects to represent datetimes in
  UTC: datetime.datetime.now(datetime.UTC).

* the decision was made to move to utc aware timestamps

* the FilterConverter addition is necessary to make the flask-admin
  work, because flask-admin doesn't know anything about the UTCDateTime
  column type


utnapischtim:
superseded by #543

PR #544, opened 2026-02-13
feat(auth): add per-account auth rate limits

slint:
- Enforce per-account limits on forgot-password, login, and send-confirmation flows using user-id limiter keys.
- Add configurable rate-limit and key-prefix settings for each protected flow.

---

## Screenshots

CleanShot 2026-02-19 at 22 43 11@2x

CleanShot 2026-02-19 at 22 44 27@2x

CleanShot 2026-02-19 at 22 45 42@2x

### DIFF
--- a/invenio_accounts/config.py
+++ b/invenio_accounts/config.py
@@ -88,6 +88,48 @@ This will be used to build an absolute URL, thus if e.g. a hostname isn't
 included, the one from the current request's context will be used.
 """
 
+ACCOUNTS_FORGOT_PASSWORD_EMAIL_RATELIMIT = None
+"""Flask-Limiter rate limit string for forgot-password requests per account.
+
+Example: ``"3 per hour"``. Disabled when ``None``.
+"""
+
+ACCOUNTS_FORGOT_PASSWORD_EMAIL_RATELIMIT_KEY_PREFIX = "accounts.fp_email"
+"""Prefix used to namespace forgot-password per-account limiter keys."""
+
+ACCOUNTS_FORGOT_PASSWORD_EMAIL_RATELIMIT_MSG = _(
+    "Too many password-reset requests for this account. Please try again later."
+)
+"""Message shown when forgot-password per-account rate limit is exceeded."""
+
+ACCOUNTS_LOGIN_RATELIMIT = None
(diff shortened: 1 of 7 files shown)


invenio_accounts/config.py
```diff
@@ -88,6 +88,48 @@
...
+
+ACCOUNTS_FORGOT_PASSWORD_EMAIL_RATELIMIT_KEY_PREFIX = "accounts.fp_email"
+"""Prefix used to namespace forgot-password per-account limiter keys."""
+
+ACCOUNTS_FORGOT_PASSWORD_EMAIL_RATELIMIT_MSG = _(
+    "Too many password-reset requests for this account. Please try again later."
+)
+"""Message shown when forgot-password per-account rate limit is exceeded."""
+
+ACCOUNTS_LOGIN_RATELIMIT = None
```
ntarocco, line 105: Do we want to have `None` as default? Or maybe a good value instead as a default?
Same for all the other endpoints.

ntarocco (approved):
Nice work!

invenio_accounts/config.py
slint: I'm not sure if Flask-Limiter is enabled by default in all instances... In `invenio-app` we always [initialize the extension](https://github.com/inveniosoftware/invenio-app/blob/153dedaba084d8b55ce98dfe5a6f003edd061e90/invenio_app/ext.py#L59), but I'm not sure if there's another config flag that actually "enables" it.

I wanted to go around this assumption by not configuring any of the limits here. This is something we could do in `invenio-app-rdm` though where we have already configured e.g. [Redis for the rate-limiting storage](https://github.com/inveniosoftware/invenio-app-rdm/blob/1e19015ac15dae33c5cb84d0175d63f9ab0cf525/invenio_app_rdm/config.py#L223).

slint:
Merging this, because we had it backported for Zenodo, and right now we kind of missed it in the latest deployment.

invenio_accounts/config.py
slint: Shelved in https://github.com/inveniosoftware/invenio-accounts/issues/549

PR #545, opened 2026-02-14
chore: compatibility webargs > 6.0.0

utnapischtim:
replacing passlib with libpass is straight forward, only replace passlib with libpass in `flask-security-fork`setup.cfg file. the from passlib statments can stay as they are!!!!

https://github.com/inveniosoftware/flask-security-fork/pull/91 needed for passlib -> libpass

the wip commit has to be removed, it is only there to show that the tests are working with this change

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,11 @@
 Changes
 =======
 
+Version v7.1.0 (released 2026-03-12)
+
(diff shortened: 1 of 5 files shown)


fenekku (approved):
Looks good, we just need to remove the github import temporarily used to show that the tests pass.

PR #557, opened 2026-06-15
feat: enforce Role id and name are equal

egabancho:
* Uses hybrid properties on both `id` and `name` to keep them in sync. Setting either field updates both column. The `name` setter raises a `ValueError` on already persisted roles to prevent unwanted primary key changes.

* There is a companion change on Invenio-App-RDM that updates old database to keep name an id in sync and all the tables that reference them, inveniosoftware/invenio-app-rdm#3459


Samk13 (approved):
This makes sense to me. In RDM, roles created via the CLI, admin UI, auth, and API already enforce `id == name`, and the companion RDM PR includes a migration script for existing instances where generated role IDs may differ from names. With that migration in place, removing the UUID default helps enforce the intended invariant.

Since renaming roles is not encouraged, and the UI now only allows editing the description, could we make the model stricter too? 

Currently `name` is immutable after persistence, but id can still be changed and will sync name as well. It seems safer to either prevent changing `id` after persistence too, or make both `id` and `name` immutable after creation.
WDYT?

egabancho:
> Currently name is immutable after persistence, but id can still be changed and will sync name as well. It seems safer to either prevent changing id after persistence too, or make both id and name immutable after creation.
WDYT?

IHMO, if you want to change `id` you do know what you're doing. 
If the role is not used anywhere, it will update it. Otherwise, the DB would reject it anyway because the FK constraints have no ON UPDATE CASCADE. PostgreSQL would raise an IntegrityError if any rows in accounts_userrole or access_actionsroles reference that `id`.

I'm fine adding the same guard for `id`, the `ValueError` in the setter is really just about giving a clearer error message earlier, before it even hits the DB.

zzacharo (approved):
It LGTM, maybe @slint wants to double check too?

invenio_accounts/models.py
```diff
@@ -88,6 +89,35 @@ class Role(db.Model, db.Timestamp, RoleMixin):
...
+
+        Raises an error if the role is already persisted, since renaming
+        would change the primary key.
+        """
+        insp = sa_inspect(self, raiseerr=False)
+        if insp is not None and insp.persistent:
+            raise ValueError("Cannot change the name of a persisted role.")
+        self._name = value
+        self._id = value
+
```
ptamarit, line 121: Which of `id` or `name` should be used by developers? Could we reflect this preference with a deprecation or in the docstring?

egabancho: Good point!
Since `id == name` always, it doesn't functionally matter. Semantically, `name` is what developers should use. `id` is just the underlying PK that happens to equal name.
I added a small comment to the column definition.

ptamarit (approved):

PR #560, opened 2026-06-23
fix(session): only set session permanent if it wasn't permanent before

max-moser:
Any write operation to the `KVSession` will mark it as modified.
Flask uses `session.modified` to check if it should add the session cookie to the response [1].
To avoid setting more session cookies on responses than necessary (e.g. because that's bad with HTTP caches), we avoid marking the session as modified unnecessarily here.

[1] https://github.com/pallets/flask/blob/3.1.3/src/flask/sessions.py#L245-L247

### DIFF
--- a/invenio_accounts/ext.py
+++ b/invenio_accounts/ext.py
@@ -338,7 +338,13 @@ class InvenioAccountsUI(InvenioAccounts):
 
         @app.before_request
         def make_session_permanent():
-            session.permanent = True
+            # `KVSession.modified` simply tracks if there were any write accesses
+            # to it, regardless of whether the value was actually changed or not.
+            # Since the `session.modified` marker is also used by Flask to determine
+            # if it should set a session cookie, and we don't want to have that
+            # set on every response, we avoid the write access here unless necessary.
+            if not session.permanent:
+                session.permanent = True
 
 
 def finalize_app(app):


invenio_accounts/ext.py
```diff
@@ -338,7 +338,13 @@ def make_session_permanent(self, app):
 
         @app.before_request
         def make_session_permanent():
-            session.permanent = True
+            # `KVSession.modified` simply tracks if there were any write accesses
+            # to it, regardless of whether the value was actually changed or not.
+            # Since the `session.modified` marker is also used by Flask to determine
+            # if it should set a `Set-Cookie` header, and we don't want to have that
+            # set on every response, we avoid the write access here unless necessary.
+            if not session.permanent:
```
utnapischtim, line 346: question: i think i don't understand that change. 
at which point in the code execution is this different than the former code?

what value can session.permanent have that it does not go into the if block?

if it is truthy, it is already True, if it is falsy it goes into and then it is True, so the same as before, but without the if block?

max-moser: So the problem that we're currently facing is that virtually every single HTTP response from Invenio gets the session cookie (`Set-Cookie` header field) set to basically the same thing.
This is bad for caching because (1) many HTTP caching solutions like NGINX or Varnish refuse to cache responses with that set out of the box, because (2) caching a response with a session cookie set gives you a nasty session stealing vulnerability if you don't explicitly strip it.
Since we want to enable more HTTP caching, this current behavior is bad for us.

Like outlined in the comment and commit message, `KVSession` tracks *any* write access to the `session` object rather than actual value modifications: https://github.com/inveniosoftware/flask-kvsession/blob/master/flask_kvsession/__init__.py#L96-L98 [1]
If `SESSION_REFRESH_EVERY_REQUEST` is set to `False`, Flask uses the `session.modified` marker to determine if it should set a session cookie for the response: https://github.com/pallets/flask/blob/3.1.3/src/flask/sessions.py#L245-L247

This change first checks the value of `session.permanent` to avoid a write access if the session is already marked as permanent, because we don't want to have the `session.modified` set due to `session.permanent` being "changed" from `True` to `True`.

[1] It looks like the API of `CallbackDict` doesn't report the key or (changed) value of the update operation; updating that API would be a larger task.

utnapischtim: ahhhh, so `permanent` is not a property but a @property which means an assignment has side effects which we don't want to have every time.

mairasalazar (approved):

invenio_accounts/ext.py
OliverGeneser: Hmm should we add a regression test confirming that the permanent session doesn't emit Set-Cookie? Seeing as the behavior depends on the side effect of KVSession.modified 🤔

OliverGeneser (approved):

invenio_accounts/ext.py
ptamarit: FYI, the regression test has been added.

